PR #4, opened 2021-06-04
feat: using a database client per api route

yasserf:


neumark:
👍  (can't approve because Im not a reviewer)

backend-common/src/services/database/database-postgres.ts
```diff
@@ -37,13 +36,16 @@ export class DatabasePostgres {
     values: Array<string | number | null | Buffer | Date> = [],
     debug?: 'debug',
   ): Promise<QueryResult<T>> {
+    if (!this.client) {
+      this.client = await this.pool.getClient()
+    }
```
neumark, line 41: Small suggestion: could we pass in this.client to callback()?

PR #501, opened 2026-04-19
Add SessionStore and unify session propagation in queue-based workflows

### LINKED ISSUE #492: Propagate session through queue-based workflow execution
## Problem

When a workflow is started via `rpc.startWorkflow()` from an authenticated HTTP request, the user session is available in the originating context. However, when the workflow runs through the queue (non-inline), the session is lost.

In `pikku-workflow-service.ts` line 834:
```typescript
session: rpcService.wire?.session,
```

This reads the session from the live `rpcService.wire`, which is only populated when the workflow runs inline. For queue-based execution, the `rpcService` wire context doesn't carry the original session.

## Expected Behavior

The session from the originating request should be serialized into `WorkflowRunWire` (or a parallel field on the workflow run record) and restored when the workflow executes through the queue.

This would allow workflows to:
- Access `session.userId`, `session.organizationId`, etc. without duplicating these in the workflow input
- Use session-based permissions (`permissions: [...]`) that check session fields
- Maintain a consistent auth context across inline and queue-based execution

## Current Workaround

Pass session fields (userId, organizationId, etc.) as explicit workflow input data alongside the session-based permission check.

## Proposed Approach

1. Extend `WorkflowRunWire` to include an optional serialized session
2. When `startWorkflow` is called, capture the current session and persist it with the run
3. When the workflow job executes from the queue, restore the session onto the wire

Related: #491 (workflow onFail callbacks)

yasserf:
## Summary

Closes #492

Introduces \`SessionStore\` as the canonical session persistence layer and refactors \`ChannelStore\` to delegate session handling. Previously, sessions for queue-based workflows had no propagation path — this branch establishes one.

### Key changes

- **New \`SessionStore\` abstraction** with in-memory, Kysely, MongoDB, and Redis implementations
- **ChannelStore refactor**: session blob removed from \`ChannelStore\`; sessions are now keyed by \`pikkuUserId\` and resolved via \`SessionStore\`
- **\`resolveSession\` helper** extracted in \`function-runner\` to unify session resolution across HTTP, CLI, MCP, scheduler, and channel runners
- **Redundant sessionStore writes skipped** when a session was loaded from the store
- **Verifier**: \`TestChannelStore\` updated for the new API; new session-store service tests added

### Affected surfaces

- \`@pikku/core\` — \`SessionStore\` interface, \`InMemorySessionStore\`, \`UserSessionService\` updates, runner integrations
- \`@pikku/kysely\`, \`@pikku/mongodb\`, \`@pikku/redis\` — session-store implementations + schema updates
- \`@pikku/cloudflare\`, \`@pikku/next\`, \`@pikku/express-server\`, \`@pikku/deploy-standalone\` — ChannelStore/session wiring updates

## Test plan

- [ ] \`yarn test\` passes across the monorepo
- [ ] Verifier tests (\`TestChannelStore\`) green
- [ ] Service tests for in-memory/Kysely/MongoDB/Redis session stores green
- [ ] Channel runner (local + serverless) still resolves sessions correctly
- [ ] Queue-based workflows propagate session to downstream steps


yasserf:
Closing — all 6 commits on this branch are already present on main as patch-equivalent commits (confirmed via `git log --cherry-mark`). The work has already landed; this PR is redundant.

PR #606, opened 2026-05-29
fix: v6 CI template failures (nextjs tsc + CLI channel single-pass)

yasserf:
## Summary

Fixes the pre-existing CI failures on `cloudflare-v6`. Root-caused three independent issues; **two are fixed and verified locally**, the third is diagnosed and in progress.

### ✅ #1 — nextjs `tsc` failure (`Templates (nextjs)`)
The nextjs template is composed by merging the `functions` template into it, pulling in `functions/scripts/` (`test-runtime.ts`, `start-mcp-http.ts`). nextjs's tsconfig type-checks `**/*.ts` under `strict`, so it compiled those helper scripts — which the functions tsconfig deliberately excludes and which reference a serverless-pruned `mcp.gen.json`. **Fix:** exclude `scripts/` in the nextjs tsconfig, matching aws-lambda/cloudflare-workers.

### ✅ #2 — CLI-over-channel 404 (`Template Functions Runtimes` × dev/standalone/serverless/cloudflare)
`pikkuCLIEntry` emits a `wireChannel()` source file (`src/wirings/cli-channel.gen.ts`) defining the CLI channel and its `cliRaw`/`cliHelp` functions. It ran *after* the channel/function/schema generators with no re-inspection afterward, so a clean single-pass build omitted it from the channel meta, function meta, and schemas — the bootstrap never registered the channel. Since `*.gen.ts` is gitignored, every fresh CI checkout hit this → `404 Channel not found: /cli/todo-cli`.

**Fix:** `pikkuCLIEntry` now reports whether it wrote a channel wiring; when it did, re-inspect and regenerate channels/functions/schemas — following the existing "Re-inspect after workflows/agents" pattern.

**Verified locally** from a clean checkout (gitignored wiring removed): single-pass `pikku all` → 2 channels, and `test:runtime dev` **and** `test:runtime standalone` both pass end-to-end (CLI channel matches; workflow/RPC/queue/trigger/MCP all green).

### 🔬 #3 — workflow timeout (`Templates (workflows-bullmq / workflows-pg-boss)`) — in progress
The workflow starts (gets a runId, status `running`) but the step never executes → 30s timeout. Both backends fail identically, so the cause is common (workflow orchestrator/step queue workers not processing), not the backend. Diagnosis ongoing.

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,1 +1,0 @@
-{"sessionId":"7fee58e3-2fa7-4e6b-802f-8a00639cb633","pid":25131,"procStart":"Tue May 12 18:31:48 2026","acquiredAt":1778664824643}
--- a/packages/addon/pikku-console/src/services/wiring.service.ts
+++ b/packages/addon/pikku-console/src/services/wiring.service.ts
@@ -396,6 +396,30 @@ function uniqueFunctionKeys(func: {
   ]
 }
 
+/**
+ * Renders a Gherkin step's attached argument (DataTable or DocString) as extra
+ * lines so a step like `When ... with:` keeps its payload. Returns a string with
+ * a leading newline to append to the step text, or '' when there's no argument.
+ * Kept as plain text (not structured) so `steps: string[]` is unchanged.
+ */
+function renderStepArgument(step: Record<string, any>): string {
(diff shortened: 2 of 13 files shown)


yasserf:
Closing — superseded by #607.

PR #695, opened 2026-06-15
fix: wire auth credentials deploy + auth codegen ordering

yasserf:
## Summary

- **fix(inspector):** credentials-only `wireAuth` now adds its source file to `state.http.files` so auth routes are imported into the bootstrap and execute at runtime. The `filterInspectorState` re-attaches the auth wiring file to any HTTP-serving unit so per-unit deploy codegen doesn't drop it.
- **feat(cli):** generate typed `defineAuth` re-export for `#pikku` import; inline `AUTH_ROUTES`; emit explicit `/auth/*` routes; split schemas; add session middleware generation; default `authFile` into scaffold dir.
- **fix(cli):** re-run secrets/credentials/variables codegen after auth gen — `pikkuAuth` generates `auth-secrets.gen.ts` concurrently with `pikkuSecrets`; re-inspecting and re-running dependent codegen after the parallel block ensures `AUTH_SECRET` appears in the deploy gate's `needs_config` check.
- **fix(kysely):** fail loud when a stored secret cannot be decrypted.

## Changeset

`@pikku/inspector` patch, `@pikku/cli` patch — see `.changeset/wire-auth-credentials-deploy.md`.

## Release Notes

* **New Features**
  * Introduced declarative `defineAuth()` configuration and discovery.
  * CLI now auto-generates separate auth **wiring** and **auth secrets** artifacts, plus optional typed auth re-exports (`authTypesFile`).
* **Bug Fixes**
  * Fixed credentials-only auth route deployments to prevent missing `/auth/*` endpoints at runtime.
  * Auth secret decryption now fails fast with clearer diagnostics when any match can’t be decrypted.
* **Improvements**
  * Added stricter auth configuration validation (including duplicate/not-exported detection) and static auth route metadata for consistent codegen.


yasserf:
Superseded by #696 (full Auth.js → Better Auth migration, single auth package, + cold-bootstrap function-types fix).

PR #739, opened 2026-06-19
Ai capability interfaces

yasserf:
## Release Notes

* **New Features**
  * Added comprehensive `pikku-emails` skill documentation with email template generation and rendering guidance.
  * Extended AI capabilities: transcription, speech synthesis, image generation, embeddings, and reranking now available.
  * Voice input/output functionality moved to core package for better integration.

* **Bug Fixes**
  * Fixed email template variable scoping to prevent unrelated template variable leakage.

* **Deprecations**
  * `@pikku/ai-voice` package deprecated; use core AI agent features instead.

* **Documentation**
  * Added detailed `pikku-emails` skill guide covering templates, variables, and integration patterns.


yasserf:
Addressed the CodeRabbit review in 72a9504. Verified each finding against the code first; all six were valid:

- **voice-input SSRF (Critical)** — added a portable guard (core runs in CF Workers, so no Node `dns` for private-IP resolution): reject loopback/private/link-local IP literals + `localhost` by default, plus an optional `allowedAudioHosts` allowlist for strict callers. Noted in a comment that this can't defend against a public hostname that resolves to a private IP / DNS rebinding — that's out of reach in core.
- **voice-input ordering (Major)** — replaced the `textParts + otherParts + Promise.all(transcriptions)` reconstruction with an in-order, sequential loop that swaps each audio part for its transcription in place. Preserves message ordering and removes the unbounded parallel downloads/transcriptions.
- **voice-output format (Major)** — `synthesizeAudio` now returns the format `generateSpeech` actually produced (`result.audio.format`), and the `audio-delta` events emit that instead of hardcoded `pcm16`. `config.format` remains the request (`outputFormat`).
- **emails layout.html (Major)** — `collectTemplateVariables` now includes `partials.layout` as a discovery root. Confirmed `layout` wraps every template at render (`serialize-emails.ts:172`) and is never referenced via `{{> layout}}`, so layout-only variables were being dropped from generated types.
- **ai-voice changeset (Major)** — bumped `patch → major`; stripping the package's exports is breaking.
- **SKILL.md fence (Minor)** — added `text` language to the directory-layout fenced block (MD040).

`@pikku/core` typechecks clean. The remaining cli tsc errors are pre-existing missing-`.pikku`-codegen noise, unrelated to these files.

PR #740, opened 2026-06-19
feat(inspector,cli): diagnostic severity model + --fail-on-* flags

yasserf:
## Why

Coded diagnostics were either `critical` (block the build) or untracked. `PKU910` (a function exposing a data-classification-branded field) was `critical`, so a single leak aborted `pikku all` → incomplete codegen → the dev server crash-loops. A security finding shouldn't stop the app from *starting* — but it should be able to block a *deploy*.

## What

A severity model for tracked diagnostics, plus CLI flags to choose what blocks.

- `InspectorLogger` gains `diagnostic({ severity, code, message })` where `severity: 'warn' | 'error' | 'critical'`. `critical(code, message)` is now thin sugar for `diagnostic({ severity: 'critical', ... })`.
- The build gate fails on `critical` **only** by default. New global flags opt into stricter gating:
  - `--fail-on-error` — also fail on `error`-severity diagnostics
  - `--fail-on-warn` — also fail on `warn` (implies `--fail-on-error`)
  - `--fail-on-critical` — always on; accepted for symmetry
- `PKU910` (data-classification leak) is reclassified `critical → error`: still printed, but no longer aborts `pikku all` / the dev server. Pass `--fail-on-error` (e.g. at deploy) to make it blocking with a recommendation to fix.

## Notes for reviewers

- The `pikku all` gate (`pikkuSummary`) receives the `CLILoggerForwarder`, not the singleton `CLILogger`. The forwarder delegates `hasBlockingDiagnostics`/`blockingSeverities` to the underlying logger, mirroring the existing `hasCriticalErrors` delegation — without this the gate throws `is not a function` on every run.
- `bootstrap.ts` / `versions-update.ts` keep using `hasCriticalErrors()` (those paths emit only criticals) — left unchanged.

## Tests

- Inspector suite: 234/234. PKU910 collector mocks now route through `diagnostic`.
- New `CLILogger` severity-gate unit tests (default critical-only / critical-always-blocks / `setFailOn(['error'])` blocks error / warn never blocks under fail-on-error).
- **New end-to-end verifier** (`verifiers/classification/src/tests/pii-gate.assert.ts`) spawns the real `pikku all` against a scaffolded project that leaks a `Private` field:
  - default → exits 0 (prints PKU910, non-blocking)
  - `--fail-on-error` → exits non-zero on the same leak
  - `--fail-on-error` on a clean project → exits 0 (no false positive)
  - Verified locally against a built CLI: 3/3, full classification suite 36/36.


yasserf:
Added a 5th commit (`b63ce93`) that CI forced out into the open: the forwarder fix here makes `CLILoggerForwarder.critical()` actually track (on main it only *printed* — line 93 was a bare `this.error(...)`). That means `pikkuVersionsUpdate`, which runs inside `pikku all`, has been **silently swallowing every contract-immutability violation**. The functions template's `versions.pikku.json` had drifted unnoticed (`onConnect@v1` input hash, plus two functions never recorded). Deleted + recreated the manifest so `pikku all` is green and idempotent again. The drift is pre-existing — `current` hash is identical on green `main` — this PR just stops hiding it.

Issue #754, opened 2026-06-21
better-auth: custom mapSession pre-empted by auto-generated stateless middleware (no opt-out)


yasserf:
## Problem

When `session.cookieCache` is enabled, the CLI auto-generates a global session middleware (`src/scaffold/auth-middleware.gen.ts`):

```ts
addHTTPMiddleware('*', [betterAuthStatelessSession()])
```

This uses the **default** session map — `{ userId: cached.user.id }` only. A project that needs extra session fields (e.g. `role`, `locale` from better-auth `additionalFields`) must register its own `betterAuthStatelessSession({ mapSession })`.

But `betterAuthStatelessSession`/`betterAuthSession` both short-circuit when a session is already set:

```ts
if (!http?.request || !setSession || session) return next()
```

The generated middleware is imported **first** in `pikku-http-wirings.gen.ts` (before user `wirings/*`), so it sets the userId-only session, and the user's custom-`mapSession` middleware then sees `session` truthy and **skips** — silently dropping `role`/`locale`.

## Impact

Any better-auth project with a custom `mapSession` (extra session fields) **cannot** enable stateless tree-shaking without regressing its session shape. There is no way to:
- feed a `mapSession` into the generated middleware, nor
- suppress the auto-registration so the project can register its own.

This blocks the ~2.5MB→~20KB per-unit tree-shaking win for exactly the apps that have non-trivial auth (custom roles/locales).

## Requested

One of:
1. Let the `pikkuBetterAuth` config declare a `mapSession` that the CLI bakes into the generated `betterAuthStatelessSession({ mapSession })`, or
2. A config/flag to suppress the auto-registration so the project owns the global session middleware, or
3. Make the generated registration run **last** (or be a no-op if a user session middleware exists).

## Workaround until then

Keep the stateful `betterAuthSession({ mapSession })` (no cookieCache) — correct sessions, but no tree-shaking.

PR #777, opened 2026-06-26
fix(auth): skip generated betterAuthSession when user registers their own

yasserf:
## Problem

The CLI's `auth.gen.ts` unconditionally wired a global `addHTTPMiddleware('*', [betterAuthSession()])` (default `{ userId }` map) on the stateful path. A project that needs a customized session bridge — `mapSession`, `impersonation`, `apiKey` — had to register a *second* global `betterAuthSession({...})`, leaving two in the chain. The generated default ran first and short-circuited (`if (session) next()`), so the user's customized one never took effect.

This is the stateful counterpart to the stateless skip already shipped in #754 (`userStatelessSession`).

## Fix

- **`@pikku/inspector`** — records `state.auth.hasUserSessionMiddleware` when it detects a user-authored **global** `betterAuthSession` registration. Route-scoped registrations (`addHTTPMiddleware('/path', …)`) and registrations inside `.gen.ts` files are ignored, so the generated middleware never self-suppresses on re-inspection.
- **`@pikku/cli`** — `serializeAuthGen` omits its own global `betterAuthSession()` from `auth.gen.ts` when `hasUserSessionMiddleware` is set. Exactly one session bridge in the chain: the user's.

Result: you register your own customized `betterAuthSession({ mapSession, impersonation })` globally and the CLI steps aside — no double-registration, no short-circuit.

## Tests

- `serialize-auth-gen.test.ts` — 3 new: drops the generated middleware when flagged, keeps handler/routes/`createAuthHandler` import, and still generates by default.
- `add-auth.test.ts` — 2 new: a global `betterAuthSession` sets the flag; a `.gen.ts` `betterAuthSession` does **not** (the self-suppression guard).

CLI 30/30 and inspector 22/22 pass.

Also updates the `pikku-better-auth` skill to document the new behavior.

## Summary by CodeRabbit

* **New Features**
  * The CLI now respects user-defined global session middleware and avoids generating a duplicate default session bridge.
  * Session wiring detection now ignores generated files and route-scoped middleware, reducing false matches during regeneration.

* **Bug Fixes**
  * Prevented generated auth middleware from overriding custom session configuration.
  * Kept stateless session behavior unchanged while improving stateful session handling.

* **Tests**
  * Added coverage for both the skip and default generation paths in auth wiring.


yasserf:
Folded into #778 (combined with the better-auth impersonation feature it enables — they ship as one cohesive change).

PR #778, opened 2026-06-26
feat(better-auth): impersonation + CLI/inspector skip generated betterAuthSession

yasserf:
Combines the impersonation **feature** with the codegen **fix** that lets you wire it cleanly. They touch disjoint packages and ship as patch bumps. (Supersedes #777, now closed.)

## 1. `@pikku/better-auth` — optional impersonation

Adds an optional `impersonation` config to `betterAuthSession` (and the stateless `betterAuthStatelessSession` variant). When configured:

- A request carrying the impersonation header (default `x-pikku-impersonate-user-id`) **and** passing the `canImpersonate` gate resolves the session as the target user via `loadUser`.
- Unknown targets fall back to the real caller with a `warn` (never a hard fail).
- Self-impersonation is a no-op.
- The header is completely inert when `impersonation` is not configured — opt-in only.

Lets an admin act as another user without a bespoke per-app middleware.

## 2. `@pikku/cli` + `@pikku/inspector` — skip the generated session bridge when the user registers their own

Before this, `auth.gen.ts` unconditionally wired a global `addHTTPMiddleware('*', [betterAuthSession()])` (default `{ userId }` map). To use a customized bridge (`mapSession`, `impersonation`, `apiKey`) you had to register a *second* global `betterAuthSession({...})` — and the generated default ran first, short-circuiting (`if (session) next()`) so yours never took effect.

- **inspector** records `state.auth.hasUserSessionMiddleware` when it sees a user-authored **global** `betterAuthSession` registration. Route-scoped registrations and registrations inside `.gen.ts` files are ignored, so the generated middleware never self-suppresses on re-inspection.
- **cli** omits its own global `betterAuthSession()` from `auth.gen.ts` when that flag is set — exactly one session bridge in the chain, the user's.

Stateful counterpart to the stateless skip in #754 (`userStatelessSession`).

## Why both together

#1 is the capability; #2 is what lets you wire it via autogeneration instead of ending up with two `betterAuthSession` in the chain. Together they're the full clean path; the fabric console admin/impersonation slice depends on both.

## Tests

- `auth-session.test.ts` — header ignored when unconfigured, gate denial falls back to caller, successful impersonation resolves target, unknown target warns + falls back, self-impersonation no-ops.
- `serialize-auth-gen.test.ts` — 3 new: drops generated middleware when flagged, keeps handler/routes/`createAuthHandler` import, still generates by default.
- `add-auth.test.ts` — 2 new: global `betterAuthSession` sets the flag; `.gen.ts` `betterAuthSession` does **not** (self-suppression guard).

CLI 30/30, inspector 22/22. Updates the `pikku-better-auth` skill.


yasserf:
@coderabbitai review

yasserf:
@coderabbitai full review

yasserf:
@coderabbitai review

PR #808, opened 2026-06-29
feat(serve): server lifecycle hooks + workflow queue wiring fixes

yasserf:
## Summary

- Add `pikkuServerLifecycle` API with `beforeStart`/`afterStart`/`beforeStop`/`afterStop` hooks, wired through the `serve` command
- Add `serve-console.ts` to split console server startup from main serve logic; add lifecycle factory detection in inspector
- Fix workflow queue worker wiring: `InMemoryWorkflowService` now defaults to `wireQueues: false` (inline-only, should never populate the shared queue worker registry); `rewireQueueWorkers()` is called on the resolved workflowService after user services are merged so wiring happens exactly once on the right service
- Add `WorkflowRunCancelledError` (registered `PikkuError`) thrown by `runToCompletion` when a workflow is cancelled, so the HTTP response includes the cancellation message
- Fix hmac-signer addon to throw `ForbiddenError` instead of plain `Error` so the missing-credential message surfaces in the HTTP response body
- Add `rewireQueueWorkers?()` to the `WorkflowService` interface (was an `any` cast)

## Test plan

- [x] All 61 e2e scenarios pass (`yarn test` in `e2e/`)
- [x] Previously failing: HMAC missing credential (×2), DSL workflow cancellation, async workflow poll-to-completion, async workflow SSE stream

## Summary by CodeRabbit

* **New Features**
  * Added support for starting the app with an optional console UI alongside the server.
  * Introduced lifecycle hooks for startup and shutdown, enabling smoother server initialization and cleanup.
  * Added clearer handling for cancelled workflows.

* **Bug Fixes**
  * Improved error handling when required credentials are missing.
  * Refined end-to-end server startup and teardown to be more reliable.


yasserf:
Superseded by #811

PR #821, opened 2026-07-01
Feat/cli audit

yasserf:
## Summary by CodeRabbit

* **New Features**
  * Added a new Security page in the console to view audit results, vulnerability details, and available dependency updates.
  * Added a command to generate and save a dependency security audit report.
  * Added update actions for vulnerable dependencies directly from the console.

* **Bug Fixes**
  * Improved handling when no audit report exists or an audit cannot be completed, showing a valid empty state instead of failing.


yasserf:
Addressed CodeRabbit's review in 761c5c3b5. Summary:

**Fixed**
- `audit.ts runBun` — added a 120s timeout + throw on launch failure / non-zero exit with no output; the `bun` branch now emits a `note` on failure so a failed run can't read as "0 advisories".
- `audit-exec.ts spawnProcess` — 5m timeout that SIGKILLs a hung child + stderr captured into the rejection.
- `update-dependency` — `bumpPackageJson` now refuses `workspace:`/`file:`/`link:`/git/url specifiers instead of clobbering them; `version` input is validated as a concrete semver.
- `agent-credential-check` — `userId` now comes from the session, not public input (closes the arbitrary-user probe).
- `useSecurityAudit` — dropped `placeholderData: null` (fixes the empty-state flash on first fetch); guarded `runSecurityAudit` `onSuccess` against a null wipe.
- `SecurityAuditView` — moved the `counts` `useMemo` above the `report.note` early return (Rules of Hooks); stable finding keys when `advisoryId` is empty.
- `get-workflow-run-history`/`-names` — descriptions now match the `MissingServiceError` throw.
- `SecurityPage` — hardcoded error string → i18n key.
- Extracted `SeverityDot`/`FindingItem`/`DependencyRow`/`CleanState` into their own `React.FC` files + a shared `security-view-utils.ts` (one component per file).

**Intentionally not changed — `process.env` forwarding (audit.ts:55, audit-exec.ts:50)**
These spread `process.env` when *spawning a subprocess* (`bun audit` / `bun install` / `pikku audit`), not to read app config. The children genuinely need the inherited environment — `PATH` to find the binary, `HOME` for bun's cache/config, and any private-registry auth — so a minimal `{PATH, HOME}` env would break installs in real projects. The `variables`-service guideline targets reading configuration inside a function, which isn't what's happening here. The blast-radius concern is also bounded now that every console RPC requires an authenticated (admin) session.

PR #860, opened 2026-07-04
feat(auth): dev quick login for the console when running locally

### LINKED ISSUE #857: Dev quick login for the console when running locally
Logging into the console during local development is tedious — every fresh session requires typing the dev admin email/password on the login screen (e.g. when opening deep links like \`/tests/userflows?id=...&runId=...\`).

Add a dev-only quick login:

- \`@pikku/better-auth\`: the generated auth catch-all handler intercepts \`<basePath>/dev/quick-login\` when explicitly enabled via env flag AND the request host is localhost. \`GET\` reports availability; \`POST\` seeds (or reuses) a dev admin user and returns a signed-in better-auth session.
- \`pikku serve\` / dev server enables the flag automatically.
- Console \`LoginScreen\` probes the endpoint for localhost servers and shows a one-click "Continue as dev admin" button.

The endpoint must never be reachable in production: it requires the explicit env opt-in (set only by the dev server) and a localhost Host header.

yasserf:
Closes #857.

Logging into the console during local dev meant typing the admin email/password on every fresh session. This adds a one-click quick login that only exists locally:

- **@pikku/better-auth** — the auth catch-all handler intercepts `<basePath>/dev/quick-login` when `PIKKU_DEV_QUICK_LOGIN` is set **and** the request host is a loopback address (defense-in-depth on top of the env opt-in). `GET` reports availability; `POST` idempotently seeds an `admin@pikku.dev` user, promotes it to `admin` via better-auth's internal adapter, and returns better-auth's own sign-in response so the session cookie is set exactly like a manual login.
- **@pikku/cli** — `pikku serve` and `pikku dev` default the flag on (`PIKKU_DEV_QUICK_LOGIN=false` opts out). Production runtimes never set it, and the endpoint additionally refuses non-loopback hosts.
- **@pikku/console** — the login screen probes the endpoint (loopback server URLs only) and shows a "Quick login as admin@pikku.dev" button that signs in with one click.

**Verification (TDD flow):**
- `packages/services/better-auth/src/dev-quick-login.test.ts` — 7 unit tests written first (failed red), pass after: flag gate, loopback gate, path fall-through, GET probe, POST seed + role promotion + sign-in, existing-user idempotency.
- `packages/console/src/lib/dev-quick-login.test.ts` — 5 unit tests for the probe/post helpers (red → green).
- End-to-end against the e2e project via `pikku serve`: `GET` → `{"enabled":true,"email":"admin@pikku.dev"}`, `POST` → 200 + session cookie, `get-session` resolves with `role: admin`; with `PIKKU_DEV_QUICK_LOGIN=false` the route falls through to better-auth's 404. In the real console UI (vite + Playwright) the button rendered and one click landed on `/overview` signed in.

Note: includes the ephemeral-port verifier commit from #859 (needed for the pre-push hook to pass locally); it will fall away on rebase once #859 merges.

## Summary by CodeRabbit

* **New Features**
  * Added a local-only one-click quick login option for the console on development servers.
  * Dev and serve commands now enable this by default, with an opt-out setting available.
  * Added support for a seeded admin sign-in flow on loopback hosts.

* **Bug Fixes**
  * Quick login is restricted to local addresses and non-production use.
  * Improved handling so the login option only appears when the server supports it.

* **Documentation**
  * Added a new localized label for the quick login button.

### DIFF
--- a/.changeset/actor-flow-approval-loop-cap.md
+++ b/.changeset/actor-flow-approval-loop-cap.md
@@ -1,0 +1,12 @@
+---
+'@pikku/core': patch
+---
+
+Bound the actor-flow approval loop (#850)
+
+`converseWithTarget` now caps suspend→approve rounds within a single target turn
+(default 16, override via `maxApprovalRounds`). A cooperative target completes
+after a handful of rounds; a buggy or uncooperative one — e.g. re-requesting a
+tool the persona keeps denying — previously could spin the inner loop forever
+without ever spending a `maxTurns` credit. Exceeding the cap now throws instead
+of hanging.
--- a/.changeset/console-run-user-flow-actors.md
+++ b/.changeset/console-run-user-flow-actors.md
@@ -1,0 +1,18 @@
+---
+'@pikku/core': patch
+'@pikku/console': patch
+---
+
+Run user flows from the console, actors and all (#850)
+
+Starting a `user-flow` workflow without explicit run actors (as the console's
+Run button does) now auto-builds HTTP actors from `USER_FLOW_ACTOR_SECRET` and
+`API_URL`: each actor signs in via the actor auth plugin — which mints the
+`actor: true` user row on first sign-in — and drives its steps over HTTP as
+that persona. When the secret or API base URL isn't configured the run simply
+proceeds without actors (with a warning) instead of failing.
+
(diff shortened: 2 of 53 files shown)


yasserf:
Superseded by #870, which combines this with #867 for a single review.

PR #867, opened 2026-07-04
feat(cli): serve the console same-origin at /console (#861)

### LINKED ISSUE #861: Serve the console at /console on the API server instead of a separate static server
`pikku serve --console <port>` currently spins up a second HTTP server that statically serves the bundled console app, which forces cross-origin auth (cookies, trustedOrigins) and a `?server=` param so the console knows where the API lives.

Serve the console from the API server itself instead, under `/console`:

- `pikku serve` / `pikku dev` mount the bundled console app (`packages/cli/console-app`) at `/console` on the same port as the API — node and bun dev servers both.
- The console is built with `base: '/console/'` (its router already honors `BASE_URL`), so assets and deep links work under the prefix.
- Same-origin means the console's default server URL is just `window.location.origin` — no `?server=`, cookies are first-party, and the dev quick login (#857) is one click away from any deep link.
- The separate static console server (`serve-console.ts`) and the `--console <port>` flag are removed (decision: same-origin only).
- e2e harness/config updated: `consoleUrl` becomes `<apiUrl>/console`.

yasserf:
Closes #861

## What

`pikku serve` and `pikku dev` now serve the Pikku Console same-origin at `/console` on the API port. The separate static console server and its `--console <port>` flag are removed.

- **`@pikku/node-http-server` + `@pikku/bun-server`**: new `staticMounts` config (`{ urlPrefix, directory, spaFallback }`) — prefix-matched static file serving with SPA fallback to `index.html`, GET/HEAD only, path-traversal protection, and correct content types.
- **`@pikku/cli`**: `serve`/`dev` mount the bundled `console-app` at `/console` when present and log `Pikku Console available at http://<host>:<port>/console`. The `--console` flag is gone. The standalone `pikku console` command serves the bundle under `/console` and redirects `/` there.
- **`@pikku/console`**: built with `base: '/console/'` (the router already derives its basename from `BASE_URL`); public assets referenced via `BASE_URL` instead of absolute `/` paths.
- **e2e**: `consoleUrl` defaults to `<apiUrl>/console`; the harness no longer spawns a second server; the obsolete `serve-console.cjs` helper is deleted.

## Why

Same-origin means auth cookies are first-party — no cross-origin cookie rewriting, no `?server=` param, no second port to manage.

## Verification

- TDD verifiers: 6 new tests per runtime (`static-mount.test.ts` in both `@pikku/node-http-server` and `@pikku/bun-server`) — written red against the unpatched servers, green after (files-under-prefix, index at mount root ± trailing slash, SPA fallback deep link, no interception outside the prefix or for POST, `%2e%2e` traversal blocked).
- Full manual flow on the e2e app: `pikku serve` on one port → `/console` deep link → login (first-party cookie) → console fully functional; `/rpc/*` and `/api/auth/*` unaffected.
- **e2e UI suite (`@console`, 28 scenarios): 14 passed / 14 failed on this branch — and the identical 14 scenarios fail at the identical steps on current `main`** (verified by rebuilding this worktree at `origin/main` and running the same suite through the old `--console` flow). The pre-existing failures are stale tests, not regressions: the addons/packages page was redesigned to cards on July 3 while `addons.steps.ts` still waits for a `table`; `workflow-ui.steps.ts` fetches `console:getWorkflowRuns` without a session and console RPCs now require auth; the User Flows scenarios expect the unmerged feat/850 redesign.


yasserf:
Superseded by #870, which combines this with #860 for a single review.

PR #925, opened 2026-07-11
feat(core,cli): outgoing webhooks — webhook service + delivery queue worker + scaffold flag

### LINKED ISSUE #922: Outgoing webhooks: webhook service + delivery queue worker + scaffold flag
Outgoing webhooks are core to a lot of apps, and Pikku has nothing for them today (every "webhook" in the repo is inbound — the gateway transport type).

Add a simple OSS building block:

- A `WebhookService` **interface** in `@pikku/core` services; the default `QueueWebhookService` implementation just enqueues delivery jobs onto a `pikku-webhooks` queue.
- A queue worker that POSTs the payload: 2xx = success, anything else throws so the queue layer retries with backoff (`attempts`/`backoff` in `JobOptions` — supported by BullMQ, pg-boss, and the in-memory dev queue).
- Per-call `retries`/`retryDelay` overrides on `send()`, defaults under `CoreConfig.webhook`.
- Optional HMAC signing: per-call `secret` → `X-Pikku-Signature: sha256=<hex>` over the body (GitHub style).
- A `scaffold.webhook` flag in `pikku.config.json` (+ `pikku enable webhook`) that generates the `wireQueueWorker` wiring, mirroring the `remoteRpc` scaffold.
- On final retry exhaustion: rely on the existing queue-runner error logging — no onFailure hook in v1.

Out of scope for v1: delivery persistence table, subscription registry, dashboard, configurable queue name.

yasserf:
Closes #922

Outgoing webhooks are core to a lot of apps; this adds a simple OSS building block: service → queue → worker with retries.

## What

- **`WebhookService` abstract class** in `@pikku/core` services (`webhook-service.ts`), owning the HMAC sign/verify, with `QueueWebhookService` as the default implementation: it JSON-serializes the payload, optionally signs it, and enqueues a delivery job onto the **`pikku-outgoing-webhooks` queue**. Apps can extend it to substitute their own implementation (direct-send, Svix, …). `QueueWebhookService` takes the queue as a **constructor dependency**, so wiring webhooks up without a queue is a type error rather than a runtime throw.
- **`pikkuWebhookWorkerFunc`** (core) POSTs the payload to the target URL with a 30s timeout; any non-2xx throws so the queue layer retries with backoff — **exponential by default**, fixed when a concrete `retryDelay` is given. Works on BullMQ, pg-boss, and the in-memory dev queue.
- **Retry policy**: per-call `retries`/`retryDelay` on `send()` override defaults in **`CoreConfig.webhook`** (`attempts = retries + 1` is always passed, mirroring the workflow service's `resolveStepJobOptions`).
- **HMAC signing**: `sha256=<hex>` over the body, sent in `CoreConfig.webhook.signatureHeader` (default `X-Pikku-Signature`). The config-level `secret` is a secret *name* resolved through the secret service; the per-call `secret` is a raw value override (e.g. per-subscriber keys from a DB). Signatures are computed at enqueue time so the key never enters the queue payload. The HMAC primitive lives in `@pikku/core/hmac` (`hmacSha256Hex`/`timingSafeStringEqual`), shared with the Slack gateway's verifier, which duplicated it.
- **Scaffold**: `scaffold.webhook: true` in pikku.config.json (or `pikku enable webhook`) generates `webhook.gen.ts` with the queue-worker wiring. It is a **boolean**, not a `PikkuScaffoldFeature` — the delivery worker exposes no endpoint, so it has no auth dimension. The payload type is an inline TS literal — a zod schema here would make the inspector import the file at codegen time, which breaks on stale deploy-unit `pikku-types` imports. The handler is inlined into `wireQueueWorker` (it is wired exactly once).

- **Bug fix (inspector)**: inlining that handler required fixing `addQueueWorker`. `addFunctions` bails on anonymous `__temp_` ids while the queue visitor renames them to a context-based `queue:<name>`, so the wiring referenced a function that was never registered and delivery failed at runtime with `Missing generated metadata for queue worker`. It now calls `ensureFunctionMetadata` (as the HTTP/MCP wirings already do) and preserves `sessionless` from the helper used — otherwise every job fails with `Authentication required`. The same gap remains in the scheduler/trigger/gateway visitors and is left for a follow-up.
- **`pikku dev`/`serve`** inject `QueueWebhookService` alongside the in-memory queue; the functions template threads it through and enables the scaffold.
- **Bug fix (CLI)**: the bootstrap missed the schemas `register.gen.ts` import when a project's only schemas came from late-generated scaffold wirings (first `pikkuSchemas` pass empty → import never pushed).

## Verifier (TDD)

New `verifiers/webhooks` workspace (added to the CI matrix): enables `scaffold.webhook`, runs codegen, and asserts against a local mock HTTP receiver:
1. delivery arrives as a POST with the exact payload and a valid signature in a **non-default, config-driven** `X-Verifier-Signature` header (config secret resolved via the secret service), asserting `X-Pikku-Signature` is absent so the header can't regress to hardcoded,
2. 500 → 500 → 200 retries until success (exactly 3 attempts),
3. `retries: 1` against a permanent 500 stops after exactly 2 attempts.

All three failed before the implementation existed. Also: 11 new unit tests in `@pikku/core` (JobOptions mapping, config/per-call precedence, HMAC round-trip, worker 2xx/non-2xx, in-memory redelivery) and serializer tests in `@pikku/cli`.

## Out of scope (v1)

Delivery persistence table, subscription registry, dashboard, configurable queue name, onFailure hook (queue runner already logs each failed attempt). Note there is **no delivery persistence**: all delivery state lives in the queue backend, so "did subscriber X receive event Y?" is unanswerable after retries are exhausted.

## Summary by CodeRabbit

* **New Features**
  * Added queued outgoing webhook delivery with configurable retries, backoff, HMAC signing, and signature verification.
  * Added persistent delivery and attempt history, including idempotent delivery tracking.
  * Added a Console Webhooks page for viewing delivery status and attempt details.
  * Added CLI scaffolding to enable and generate webhook workers.

* **Bug Fixes**
  * Improved inline wiring metadata detection and preserved sessionless behavior.
  * Fixed late-generated schema registration during bootstrap.

* **Tests**
  * Added comprehensive webhook delivery, retry, signing, persistence, and end-to-end coverage.


packages/cli/src/functions/wirings/webhook/serialize-webhook.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const webhookDeliveryHandler = pikkuSessionlessFunc<
+  { url: string, event?: string, body: string, headers: Record<string, string> },
+  void
+>({
+  tags: ['pikku'],
+  func: async (services, data) => pikkuWebhookWorkerFunc(services, data),
+})
+
+wireQueueWorker({
```
yasserf, line 22: why not just inline the function?

packages/core/src/services/queue-webhook-service.ts
```diff
@@ -0,0 +1,112 @@
...
+    const backoff =
+      retryDelay !== undefined && retryDelay !== 'exponential'
+        ? { type: 'fixed', delay: getDurationInMilliseconds(retryDelay) }
+        : retries > 0 || retryDelay === 'exponential'
+          ? 'exponential'
+          : undefined
+    return { attempts: retries + 1, ...(backoff ? { backoff } : {}) }
+  }
+
+  private verifyQueueService(): QueueService {
```
yasserf: this shouldb done via cli as well to catch earlier

packages/core/src/services/webhook-service.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export interface WebhookJobData {
+  url: string
+  event?: string
+  body: string
+  headers: Record<string, string>
+}
+
+export const PIKKU_WEBHOOK_QUEUE_NAME = 'pikku-webhooks'
```
yasserf: outgoing webhooks*

packages/core/src/services/webhook-signature.ts
```diff
@@ -0,0 +1,33 @@
+import { createHmac, timingSafeEqual } from 'node:crypto'
+
+/**
+ * Sign a webhook body with HMAC-SHA256.
+ *
+ * @param secret - The signing key shared with the receiver
+ * @param body - The raw request body string
+ * @returns The `X-Pikku-Signature` header value (e.g., `sha256=abc123...`)
+ */
+export function signWebhookBody(secret: string, body: string): string {
```
yasserf: why isn't this on the webhook interface? should we make it an abstract class? I dont like functions inside of servcies..

templates/functions/pikku.config.json
```diff
@@ -30,6 +30,7 @@
     "console": "auth",
     "events": "no-auth",
     "remoteRpc": "no-auth",
+    "webhook": "no-auth",
```
yasserf: should be true, no-auth makes no sense

packages/core/src/services/webhook-signature.ts
yasserf: this duplicated/

yasserf: also signature header should be passed in from pikku.cponfig.json

packages/cli/src/functions/wirings/webhook/serialize-webhook.ts
yasserf: Done — inlined. It needed an inspector fix to actually work, though.

Inlining alone produced a scaffold that codegen accepted and that failed at runtime with `Missing generated metadata for queue worker 'pikku-outgoing-webhooks'`. `addFunctions` bails on anonymous funcs (`if (pikkuFuncId.startsWith('__temp_')) return`), while `addQueueWorker` renames that same id to a context-based `queue:<name>` — so the wiring pointed at a function that was never registered. HTTP and MCP already call `ensureFunctionMetadata` for exactly this; queue/scheduler/trigger/gateway never did.

`addQueueWorker` now registers it, and also preserves `sessionless` from the helper used (`expression.text !== 'pikkuFunc'`, same rule as `addFunctions`) — without that the stub is treated as session-required and every job fails with `ForbiddenError: Authentication required`. Covered by a new `add-queue-worker.test.ts` (fails without the fix) and the verifier is green with the inlined scaffold.

I left the same gap in scheduler/trigger/gateway alone — same one-line shape, but out of scope here and untested. Happy to file it.

packages/core/src/services/webhook-service.ts
yasserf: Renamed the queue to `pikku-outgoing-webhooks` (const `PIKKU_OUTGOING_WEBHOOK_QUEUE_NAME`) so an incoming-webhook feature can't collide with it later, and reworded the prose that said "webhooks" where it meant outgoing ones.

templates/functions/pikku.config.json
yasserf: Done — `"webhook": true`. Also changed the type: `scaffold.webhook` is now `boolean` rather than `PikkuScaffoldFeature`, since a delivery worker exposes no endpoint and has nothing to authenticate. That also drops the meaningless `pikku enable webhook --no-auth` flag.

packages/core/src/services/webhook-signature.ts
yasserf: Done — `WebhookService` is now an abstract class that owns the signing:

```ts
export abstract class WebhookService {
  abstract send(input: SendWebhookInput): Promise<SendWebhookResult>
  protected sign(secret: string, body: string): string
  public verify(secret: string, signature: string, body: string): boolean
}
export class QueueWebhookService extends WebhookService { ... }
```

`webhook-signature.ts` is deleted, so there are no loose functions under `services/` any more. `sign` is protected (implementations use it); `verify` is public for receivers.

yasserf: Yes — `packages/gateways/slack/src/slack-signature.ts`. Same HMAC-SHA256 + `timingSafeEqual`-in-a-try shape, differing only in the prefix (`v0=` vs `sha256=`) and Slack's timestamp replay guard.

Extracted the primitive to `@pikku/core/hmac` (`hmacSha256Hex` + `timingSafeStringEqual`); both the Slack verifier and `WebhookService.sign` now call it, so the scheme stays per-caller but the crypto exists once. Slack's behaviour and public signature are unchanged.

yasserf: Done, with one correction on the location.

It's `CoreConfig.webhook.signatureHeader` (defaulting to `X-Pikku-Signature`), not `pikku.config.json`. `pikku.config.json` is the CLI's codegen config — it never reaches the runtime, and signing happens at enqueue time inside `QueueWebhookService.send()`. `CoreConfig.webhook` is where `retries`/`retryDelay`/`secret` already live, so the header sits with them.

The verifier now configures a non-default `X-Verifier-Signature` and asserts both that it is used and that `X-Pikku-Signature` is absent, so this can't regress to hardcoded.

packages/core/src/services/queue-webhook-service.ts
yasserf: Caught at compile time instead — `verifyQueueService()` is gone entirely.

`QueueWebhookService` now takes the queue as a constructor dependency rather than reaching for `getSingletonServices().queueService`, so a project that wires up outgoing webhooks without a queue is a type error in `services.ts` (surfaced by `pikku all --tsc`), not a throw on the first `send()`:

```ts
export class QueueWebhookService extends WebhookService {
  constructor(private queueService: QueueService) { super() }
}
```

I avoided the other route — forcing `queueService` into `RequiredSingletonServices` when `scaffold.webhook` is on. It breaks the template, which passes `existingServices?.queueService` (possibly undefined), so `Required<>` would fail typecheck on a project that is actually fine. The template now reads honestly: no queue means no webhook service.

```ts
webhookService:
  existingServices?.webhookService ??
  (queueService ? new QueueWebhookService(queueService) : undefined),
```

packages/core/src/services/queue-webhook-service.ts
```diff
@@ -0,0 +1,112 @@
...
+import {
+  DEFAULT_WEBHOOK_SIGNATURE_HEADER,
+  PIKKU_OUTGOING_WEBHOOK_QUEUE_NAME,
+  type SendWebhookInput,
+  type SendWebhookResult,
+  type WebhookJobData,
+  WebhookService,
+} from './webhook-service.js'
+
+export const DEFAULT_WEBHOOK_RETRIES = 3
```
yasserf: should be in pikku.config.ts or config.ts

packages/core/src/services/queue-webhook-service.ts
```diff
@@ -0,0 +1,112 @@
...
+    const body = JSON.stringify(input.data)
+    const headers: Record<string, string> = {
+      'Content-Type': 'application/json',
+      ...(input.event ? { 'X-Pikku-Event': input.event } : {}),
+      ...input.headers,
+    }
+
+    let secret = input.secret
+    if (secret === undefined && webhookConfig?.secret) {
+      secret = await services.secrets.getSecret(webhookConfig.secret)
```
yasserf, line 68: we should cache this likely.

packages/core/src/services/queue-webhook-service.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+    if (secret) {
```
yasserf, line 75: if secret doesn't exist it shoudl def warn or error

packages/core/src/services/queue-webhook-service.ts
yasserf: Done in 93725d33f. There's no `config.ts` in core — the existing convention is an exported `DEFAULT_*` const (e.g. `DEFAULT_STEP_RETRIES` in `pikku-workflow-service.ts`). I moved `DEFAULT_WEBHOOK_RETRIES` out of the queue impl and into `webhook-service.ts`, co-located with the `WebhookServiceConfig` interface and `DEFAULT_WEBHOOK_SIGNATURE_HEADER`, so it reads as the documented fallback for `config.webhook.retries`. It's still overridable per-call and via `config.webhook.retries` — the const is only the last-resort default.

packages/core/src/services/queue-webhook-service.ts
yasserf: Done in 93725d33f. Config-secret resolution now goes through `resolveConfigSecret()`, which caches the resolved value in a per-instance `Map` keyed by secret name, so repeated `send()` calls don't re-hit the secret service. (A per-call `input.secret` is a raw key, not a name, so it's never resolved or cached.)

packages/core/src/services/queue-webhook-service.ts
yasserf: Done in 93725d33f. If `config.webhook.secret` names a secret that resolves to nothing, `resolveConfigSecret()` now logs `logger.error(... will be sent UNSIGNED)` and returns undefined (and doesn't cache the miss, so fixing the secret takes effect without a restart). I went with error-level rather than throwing so a misconfig doesn't hard-break every delivery, but it's loud — an unsigned webhook when you configured signing is a real problem.

packages/core/src/services/queue-webhook-service.ts
yasserf: Moved the caching into the secret service, per your follow-up — app code stays naive and just reads the latest. `TypedSecretService` now caches successful `getSecret` reads in-process (invalidated on `setSecret`/`deleteSecret`, `getSecrets` shares the same cache), so `QueueWebhookService` dropped its own per-instance cache and just calls `getSecret`. No TTL, so an out-of-band-rotated secret won't refresh until restart — tracked in #964. (3c7a8f2c2)

PR #944, opened 2026-07-16
chore(cli): bundle two skills, fix frontmatter 46 skills pi.dev silently dropped, add --agent pi

yasserf:
Bundles two new skills into `@pikku/cli`, plus test coverage for the skills corpus and the `skills install` command (which had none). That corpus test then found that **46 of the 61 bundled skills were never loading in pi.dev**, so the fix for that is here too, along with a `pi` install target.

## The skills

**`pikku-software-archaeology`** — reverse-engineers an existing repository into a Product Blueprint: a `.knowledge/` directory of schema-validated JSON plus a human-readable `blueprint.md`, recovering domains, entities, commands, queries, events, policies, workflows and invariants so an organically-grown app can be rebuilt cleanly (e.g. as a Pikku app). It's language-agnostic by design — the model reads the code rather than relying on per-language AST tooling — and ships `scripts/validate.mjs` (node, no deps) to enforce the output contract, plus `references/pikku-mapping.md` covering how a blueprint maps onto Pikku primitives.

**`pikku-product-second-opinion`** — consumes that blueprint and produces a plain-language report for a non-technical owner: how the app they hold actually works, what's solid, what's holding them back (with business impact and effort), and an argued case for a better design.

## Decisions worth reviewing

- **Both carry the `pikku-` prefix**, so all 61 bundled skills follow one convention. Neither skill is actually Pikku-specific (`pikku-software-archaeology` is framework-agnostic; `pikku-product-second-opinion` is business-facing), so the prefix is a bit meta — consistency and grouping in `pikku skills list` won over literal accuracy. The cross-references in both SKILL.md **descriptions** were renamed too, not just the directories, since the description is what routes a skill.
- **`pikku-software-archaeology/example/` (a 176K fixture Express app) was dropped** to keep the published tarball lean; `84K` instead of `260K`. Its `scripts/` and `references/` are kept, since the skill invokes the validator and maps blueprints onto Pikku primitives. `pikku-product-second-opinion/example/sample-report.md` is kept — it's 4K and its SKILL.md cites it as the section-shape reference.
- **Both `README.md` files are kept**, though no existing skill has one. Nothing loads them at runtime and they carry maintainer-facing design rationale (notably *why* there is deliberately no AST scanner).
- **Both are tagged `installGroups: [fabric]`**, so `pikku skills install --fabric` picks them up alongside `pikku-fabric`, `pikku-deploy-cloudflare` and `pikku-schema-cfworker` (3 → 5 skills). `--core` stays at 30 and excludes both.

## 46 of 61 skills were silently not loading in pi.dev

The corpus test needed a YAML parser, which forced the question of *whose* parser — and the answer is the agent that actually reads these files. pi.dev parses SKILL.md frontmatter with eemeli `yaml` (`yaml@2.9.0`, via `src/utils/frontmatter.ts`). Its `loadSkillFromFile` catches a parse error, files a **warning** diagnostic and returns `skill: null` — the skill is dropped with no loud failure.

Most of our descriptions are multi-line single-quoted scalars with continuation lines at column 0. That's **invalid per the YAML spec** and lenient parsers (js-yaml@4, which Claude Code uses) accept it, so it looked fine. eemeli rejects it with `Missing closing 'quote`. Net effect: **15 of 61 skills reached pi**; `pikku-http`, `pikku-queue`, `pikku-workflow` and `pikku-better-auth` were among the 46 that didn't.

All 46 descriptions are normalized onto `>-` folded block scalars. The rewrite was mechanical and verified rather than hand-edited: js-yaml@4's parse of each file was taken as ground truth for the *intended* description, and the rewrite was only accepted if eemeli then produced a **byte-identical description and an identical frontmatter object** — the description drives skill routing, so a subtly reworded one is a silent regression. Already-valid skills were left untouched. 46 rewritten, 15 already clean, 0 failures. Simulating pi's loader over the result: **61/61 loaded, 0 dropped, 0 warnings** (was 15/61).

The corpus test now parses with `yaml` — already a direct dependency of `@pikku/cli`, no new dep — and enforces pi's 1024-char description cap. Validating with the strict parser is the whole point: the lenient one accepts frontmatter that ships broken.

## `pikku skills install --agent pi`

Skills land in `.pi/skills/`, mirroring pi's own resolution (`core/skills.ts` scans `<cwd>/<CONFIG_DIR_NAME>/skills`; `CONFIG_DIR_NAME` defaults to `.pi`). `skills list` reports skills installed there, and `--agent pi` composes with `--only` / `--core` / `--fabric` / `--update` like the other targets.

Per-agent paths now come from one `AGENT_SKILL_DIRS` map instead of ternaries spread across the file, so the supported-agent list, install root, destination label and installed-skill scan can't drift apart. Adding a target is one entry.

This also fixes the `--agent` help, which advertised `codex | gemini` — both of which error at runtime.

## Tests

`skills.ts` had no coverage at all. Two suites, 19 tests:

**Corpus lint** — data-driven over all 61 skills, so a malformed skill fails CI instead of misbehaving silently in an agent session: SKILL.md present; frontmatter parses under **pi's own parser** with a usable name/description within pi's length cap; `name` matches its directory; names unique; `installGroups` drawn from the known set (a typo'd group silently makes a skill unreachable by any flag); and **every relative path a skill cites resolves on disk** — this one catches a doc pointing at a file trimmed from the tarball, which is a real mistake I made and it caught.

**Install behaviour** — drives the exported command against a temp cwd rather than extracting private helpers: `--only` installs exactly the named skills and copies the whole directory; unknown `--only` name or unsupported `--agent` exits 1 and installs nothing; `--agent opencode` targets `.opencode/skills` and `--agent pi` targets `.pi/skills`; a repeat install leaves an edited skill alone while `--update` overwrites; and `--core` installs the core-tagged **subset**.

That last subset assertion is the point: `getWantedSkills` falls back to "install everything" when no skill matches the requested group, so an untagged corpus would silently turn `--core` into install-all.

`--fabric` and `--core --fabric` (union) are covered too, plus an explicit pin of fabric group membership.

That pin matters more than it looks. The subset tests derive their expected set *from the corpus*, which makes them vacuous for membership — dropping a skill's `fabric` tag shrinks both sides and still passes. Verified by mutation: removing the tag from `pikku-software-archaeology` failed nothing until the pin existed, and fails loudly now. Adding a skill to the Fabric set means updating `FABRIC_SKILLS` on purpose.

The rest is mutation-verified too — neutering the skip-unless-update guard fails exactly the update test; making the group filter fall through to `allEntries` fails four, including the `--core` subset check; pointing the `pi` target at `.claude` fails exactly the three pi tests; and restoring the old frontmatter dialect on one skill fails the parse test.


yasserf:
Thanks — worked through all 6. Fixed 3, declining 3 with reasoning. Applied in ee0eb78a0.

## Fixed

**Restrict sorting to unordered identity collections** — agreed, and this was the most serious of the six. `workflows[].steps` is an ordered `string[]` in the schema, so "sort every array by `name`" would alphabetize a workflow's steps and rewrite the very behavior the skill just extracted. Scoped sorting to unordered identity collections and called out sequence-bearing arrays explicitly, in both `SKILL.md` and `README.md`.

**Use fields that exist in `migration.json`** — correct. `migration.json` is `{mappings, dropped, decisionsNeeded}`; there is no `keep-list`, so the skill was sending the reader after an input that cannot exist. Both mentions (L68 and the L129 objection table) now cite `mappings[]` (what survives, each with its `recommendation`) and `dropped[]`.

**Require line locations for evidence** — correct and safe. `SKILL.md` requires file-and-line to justify `confidence: "high"`, but the schema only required `file`, so a file-only citation could back a high-confidence claim and still validate. `lines` is now required. I checked this against the skill's worked example before applying: all 172 evidence entries already carry `lines`, so it tightens enforcement without breaking real generator output. Verified both ways — the example still validates at `0 error(s)`, and stripping `lines` from one entry now fails.

## Declining

**Treat TanStack Start as pre-1.0** — this one is factually out of date. TanStack Start shipped **v1.0 stable in March 2026**; the release-candidate status was announced in September 2025. Applying the suggestion would *introduce* an error into the skill. The existing framing (ecosystem size and hiring pool, explicitly "not a viability" concern) is accurate as written, and the skill's job here is honest both-sides tradeoffs — overstating risk fails that as badly as understating it.

**Do not generate inline confidence comments in TypeScript** — the "no new inline comments" guideline governs code written *in this repo*. `pikku-mapping.md` describes what a generator emits into a **user's newly generated application**, and `// BLUEPRINT-MEDIUM-CONFIDENCE: <evidence>` is a deliberate review marker placed at the point of uncertainty — closer to the guideline's "explain why when it isn't obvious" exception than to redundant narration. A separate manifest would move the signal away from the code it's about, which is the opposite of what a reviewer needs. Happy to revisit if you disagree.

**Actually parse the complete frontmatter block** — the suggestion is to use "the production metadata parser", but there isn't one: `skills.ts` parses frontmatter with regex itself (`parseInstallGroups`), so the test deliberately mirrors production behavior rather than holding it to a stricter standard than the code it covers. `yaml` is only *transitively* resolvable and isn't a declared `@pikku/cli` dependency, so reaching for it would add a hidden dep. I agree the test name slightly overclaims. Happy to add `yaml` as a devDep and parse strictly if that's wanted — it would catch a malformed-frontmatter skill shipping broken.

On the ast-grep ReDoS warning for `new RegExp` from variable input: not reachable here — `field()` is only ever called with three hardcoded keys (`name`, `description`, `installGroups`), never user input.

yasserf:
Correction on my "Actually parse the complete frontmatter block" reply — I gave a wrong reason, and the real one is more interesting.

**What I got wrong:** I claimed `yaml` was only transitively resolvable and adding it would introduce a hidden dep. That's false. `yaml@^2.8.2` is already a **direct dependency** of `@pikku/cli` and is imported today in `src/functions/wirings/http/pikku-command-openapi.ts`. There was no dep to add and no lockfile risk. My check ran from the wrong directory, printed nothing, and I read that silence as "not declared."

**So I implemented the suggestion — and it failed 46 of the 61 bundled skills.**

The corpus writes descriptions as multi-line single-quoted scalars with continuation lines at column 0:

```yaml
description: 'Use when creating or consuming reusable function packages (addons) in Pikku. ...
TRIGGER when: code uses wireAddon/addon()/pikkuAddonServices, ...
DO NOT TRIGGER when: user asks about internal function composition ...'
```

Per spec a single-quoted scalar may span lines, but continuations must be indented past the key; at column 0 the scalar terminates and you get "Missing closing 'quote". **But the two major parsers disagree here:**

| parser | result on `pikku-addon` frontmatter |
|---|---|
| `yaml` (eemeli, the CLI's dep) | REJECTS — `Missing closing 'quote at line 2, column 223` |
| `js-yaml` | parses fine — 510-char description |

So the 46 skills are not broken; they parse under the lenient reading that frontmatter tooling generally uses. A strict `yaml`-based assertion would have produced 46 false failures against skills that work, in a PR that touches none of them.

**Net:** declining still, but on the real grounds — "use a real parser" is only meaningful if it's the *same* parser the consumer uses, and here the two disagree on precisely this construct. The regex check deliberately mirrors `skills.ts`'s own `parseInstallGroups`, so the test holds the corpus to exactly the contract the CLI enforces. Worth noting both skills added in this PR *do* parse strictly clean under both parsers.

The wider question — whether the corpus should be normalized onto a dialect every parser accepts (indenting continuations, or using a `>-` block scalar) — is real but belongs in its own issue, not here.

yasserf:
## Retraction: my TanStack decline was wrong, CodeRabbit was right

Earlier I declined the "treat TanStack Start as pre-1.0" finding and said it was ~10 months out of date because **v1.0 shipped in March 2026**. That is not true. I did not verify it before asserting it, and I shouldn't have posted it as a correction. Fixed in c14066480.

The actual current status, from tanstack.com/start today:

> **RC status** — Feature-complete, still listening. The RC API is considered stable and preparing for 1.0. For production apps, lock dependencies to a specific version and keep up with the final feedback, docs polish, and last-mile fixes.

So there is no stable 1.0, and the vendor itself tells production users to pin — which is exactly what the finding said.

**What misled me:** `@tanstack/react-start` is on `1.168.28`, and has been on 1.x since Feb 2025. That looks like "well past 1.0" but means nothing here — Start tracks the TanStack Router version line, so its major has never encoded its own release stage. I inferred from the version number instead of checking the docs.

**Why this one mattered more than a normal miss:** the paragraph it was defending told a non-technical owner the framework was "production-proven at real scale" and that the only cost was familiarity and hiring — "not whether the framework will be around". That sits directly beneath the skill's own rule that *"Overstating cons is as dishonest as hiding them."* Hiding one is the same failure pointed the other way, and this report exists precisely for someone who cannot go check the claim themselves. Declining the finding would have shipped that.

The section now:
- separates the **mature ecosystem** (Query/Router/Table — genuinely battle-tested) from the **pre-1.0 framework** wrapping them, which is the distinction the old text blurred;
- states the real cost — pin an exact version, budget for upgrade work into 1.0 — instead of familiarity/hiring;
- drops the "production-proven at real scale" framing and the unverified vendor anecdote that was its only support;
- tells the agent to **re-check the release stage at write-up time** rather than trusting the example, since release stages move and a stale maturity claim is exactly what caused this.

## Second thread: `pikku-mapping.md` medium-confidence marker — also valid, also fixed

`// BLUEPRINT-MEDIUM-CONFIDENCE: <evidence summary>` told the generator to emit inline comments, against the repo's "do not add new inline comments" guideline. Medium confidence now surfaces in the pre-generation report instead, which is where `confidence: low` already goes — so all three tiers route through one review surface and the generated code stays clean.

Both threads can be resolved.

PR #952, opened 2026-07-16
feat(core): workflow.approval() — a return-valued, expiring human-in-the-loop gate

### LINKED ISSUE #947: Workflow approval: return-valued, expiring human-in-the-loop step
## Problem

`workflow.suspend(reason)` is a **one-shot gate, not an approval**. Three gaps make human-in-the-loop approval impossible to express today:

1. **No decision.** `suspendStep()` calls `setStepResult(...)` — which sets the step `succeeded` — and *then* throws. On replay the `status === 'succeeded'` guard fires and the workflow walks straight past. Whoever calls `resumeWorkflow(runId)` resumes it, always forward. There is no approve-vs-deny and no record of who decided.
2. **No return value.** `WorkflowWireSuspend = (reason: string) => Promise<void>`, so a "send a form, get typed data back" step cannot be expressed at all.
3. **No expiry.** A suspended run waits forever. `WorkflowStepOptions` already carries a `// Future: timeout` note.

Consequence: every approval flow in `verifiers/workflows/` is hand-rolled from `workflow.do` + `sleep`, and `multi-level-approval.workflow.ts` mocks the decision outright ("mock - in reality would check review status").

## Proposal

A return-valued, expiring approval step:

```ts
const decision = await workflow.approval('approve-invoice', {
  schema: ApprovalDecision,   // zod VALUE — validated at the resume boundary
  expiry: '3d',
})
// decision: { status: 'approved', data } | { status: 'expired' }
```

Plus a public `approveStep(runId, reason, value)` on `WorkflowService` and a generated
`POST /workflow/:name/approve` route.

### Deliberately out of scope

Assignee, notification medium (Slack/email), reminders, and escalation-to-another-user are **app
code**, not framework surface. Pikku owns only what can't be built on top: the durable pause, the
typed payload, and the expiry.

## Design constraints found while scoping

- **Schema must be a value, not a generic.** The resume payload arrives over HTTP from an untrusted
  caller; a TS generic is erased and validates nothing. The generated `TypedWorkflow` also has no
  schema source for an approval payload (`do` keys off `FlattenedRPCMap`; approval has no map), so a
  caller-supplied generic would be unvalidated at the boundary. Schema-as-value also keeps us on the
  supported side of the generics-vs-`input`/`output` rule in CLAUDE.md.
- **`succeeded` is doing double duty.** It is both the idempotency gate and (for approval) the
  value-delivery channel. These conflict: the result slot must stay empty until an external approve
  call fills it, or replay returns the sentinel instead of the decision. This is the core rework.
- **`suspend` emits zero graph nodes.** `convert-dsl-to-graph.ts` has `default: return []` and there
  is no `'suspend'` in `FlowType`. Fine for a void gate; an approval whose value feeds a later step
  must be a real node or `$ref` resolution and `graphHash` won't see it.
- **Extraction is on the wrong AST path.** `isWorkflowSuspendCall` is only consulted from the
  expression-statement branch. `const d = await workflow.approval(...)` is a *variable declaration*,
  so it needs the assignment path plus `outputVars`/`arrayVars` bookkeeping (today only
  `workflow.do()` participates).
- **`reason` is the step's durable identity** (`__workflow_suspend:${reason}`, deduped by
  `nextStepKey`). Two approvals sharing a reason collapse onto one row and one result — harmless for
  void, wrong for data.
- **Storage needs no migration.** `updateRunState`/`getRunState` is a schemaless JSON blob on all
  five backends (in-memory, Kysely `workflow_runs.state`, Redis, Mongo, Cloudflare DO). This matters:
  there are no SQL migrations in the repo — schema is built imperatively with
  `.createTable(...).ifNotExists()`, so a new column would only apply to fresh databases and silently
  skip existing ones.
- **Expiry must be durable and idempotent.** Route it through the existing `sleep` orchestrator path,
  not `setTimeout` (a 3-day approval outlives deploys). Enqueued jobs can't be reliably retracted, so
  the timer fires unconditionally and no-ops if the decision already landed.
- **Do not reuse the AI-agent approval stack.** `AgentRunState` is welded to the agent domain
  (required `agentName`/`threadId`/`resourceId`/token usage), every `PendingApproval` keys on
  `toolCallId`, and `findRunByToolCallId` full-scans all runs. It also looks incomplete:
  `resolveApproval(toolCallId, status)` drops `status` on the floor and never persists
  approved-vs-denied, and `AIRunTable` doesn't declare `pendingApprovals` (hence the `as any` cast) —
  the DDL and type row have drifted.

## Open question

What does expiry do when it fires — throw, or return a discriminated union? Union is proposed above:
it forces callers to handle the case and makes "skip and continue" trivial.

## Affected layers

| Layer | Work |
|---|---|
| Storage | none (run-state blob) |
| Core types + 3 barrels | meta type, wire signature, union member |
| Inspector | pattern, extractor, assignment-path dispatch, `FlowType`, graph node, plan step |
| Workflow service | state-machine rework; public `approveStep()` |
| CLI codegen | `workflowApproveCaller`, `wireHTTPRoutes` entry, analyzer route + skip-list prefix |
| Tests | service, plan, routes, types (no extractor test exists today — net-new) |
| Console | no "awaiting approval" rendering — separate follow-up |

yasserf:
Closes #947.

## Why

`workflow.suspend(reason)` looks like an approval primitive but isn't one. It marks its own step
`succeeded` and *then* throws:

```ts
await this.setStepResult(stepState.stepId, { reason, suspendedAt })  // ← status = 'succeeded'
throw new WorkflowSuspendedException(runId, reason)
```

On replay the `status === 'succeeded'` guard fires and the workflow walks straight past. So whoever
calls `resumeWorkflow(runId)` resumes it, always forward — there is no approve-vs-deny, no decision
handed back, and no deadline. That's why every approval flow in `verifiers/workflows/` is hand-rolled
from `do()` + `sleep()`, and why `multi-level-approval.workflow.ts` mocks the decision outright
("mock - in reality would check review status").

## What

```ts
const decision = await workflow.approval('Approve invoice', {
  schema: ApprovalDecision, // any standard-schema value (zod, valibot, arktype)
  expiry: '3d',
})
if (decision.status === 'expired') return { skipped: true }
// decision.data is typed from the schema
```

Decisions land via `workflowService.approveStep(runId, reason, decision)` or the generated
`POST /workflow/:workflowName/approve/:runId`.

## Three decisions worth reviewing

**The schema is a value, not a type generic.** The payload arrives from an untrusted caller over the
approve wire; a generic is erased at compile time and would validate nothing. The generated
`TypedWorkflow` also has no schema source to key an approval's return type off (`do` uses
`FlattenedRPCMap`; an approval has no map), so a caller-supplied generic would be unvalidated at the
boundary. Schema-as-value also stays on the supported side of the generics-vs-`input`/`output` rule
in CLAUDE.md, and core keeps its zero-zod dependency — `StandardSchemaV1` is the only contract.

**Validation runs on replay, inside the workflow body.** `approveStep` is called from an HTTP route,
a different process from the one running the workflow, so the schema value simply isn't in scope
there — and it can't be serialized into the step row. The body *is* the only place it exists. An
invalid payload therefore re-closes the gate and records the issues in run state, rather than failing
the run: otherwise any caller could kill a workflow with a malformed body.

**Expiry is evaluated against a recorded deadline, not by a timer firing.** The wake-up is a delayed
orchestrator pass — deliberately *not* `scheduleSleep`, which resolves the step it's handed and would
resolve the gate itself (this was a real bug in development: the gate came back `null`). A duplicate,
late, or dropped wake all produce the same answer; losing it costs liveness, never correctness.

## Out of scope, deliberately

Assignee, notification medium (Slack/email), reminders, and escalation-to-another-user are app code.
Pikku owns only what can't be built on top: the durable pause, the typed payload, the expiry. Approver
identity isn't modelled either — gate the approve route with your own auth.

## Notes

- **No schema migration.** State lives in the existing run-state blob, which every backend
  (in-memory, Kysely, Redis, Mongo, Cloudflare DO) already stores as schemaless JSON. This matters
  more than it sounds: there are no SQL migrations in the repo — schema is built imperatively with
  `.createTable(...).ifNotExists()`, so a new column would only apply to fresh databases and silently
  skip existing ones.
- **`approval` is a real graph node**, unlike `suspend` — which emits none, falling into
  `convert-dsl-to-graph`'s `default: return []`. Harmless for a void gate; an approval's value can be
  referenced downstream, so `$ref` resolution and `graphHash` have to see it.
- **The agent approval stack was not reused.** `AgentRunState` is welded to the agent domain
  (required `agentName`/`threadId`/`resourceId`/token usage), every `PendingApproval` keys on
  `toolCallId`, and `findRunByToolCallId` full-scans all runs. It also looks incomplete —
  `resolveApproval(toolCallId, status)` drops `status` and never persists approved-vs-denied, and
  `AIRunTable` doesn't declare `pendingApprovals` (hence its `as any`). Worth a separate look.

## Known limitation

`reason` is the gate's durable identity and `nextStepKey` dedupes by name, so an approval reached
more than once under the same reason (e.g. in a loop) gets `#N`-suffixed step rows that
`approveStep(runId, reason, ...)` can't currently target. It addresses the first reach.

## Tests

TDD throughout — each test failed before its implementation. Five new core tests cover the behaviour
`suspend` can't express:

- resuming **without** a decision must not fall through the gate
- `approveStep` records a decision the workflow then returns
- an invalid payload leaves the gate closed and is legible in run state
- an expired approval returns `{ status: 'expired' }` instead of hanging
- a decision that landed before expiry wins over the expiry

Green: core 1088, inspector 324, cli 324, 0 failures. `tsc` clean across core, inspector, cli,
cloudflare, kysely, redis, mongodb.

## Follow-ups

- Console has no "awaiting approval" rendering — a suspended run is still stored as a fake error row
  (`code: 'WORKFLOW_SUSPENDED'`), so the UI can't distinguish it from a failure without sniffing that
  string.
- Immediate approver feedback (validating at `approveStep` rather than on replay) needs the inspector
  to emit a `schemaName` into the named-schema registry.


yasserf:
@coderabbitai review

PR #960, opened 2026-07-17
feat: scopes — static-checked authorization scopes on pikkuFunc

### LINKED ISSUE #956: feat(core): scopes — static-checked authorization scopes on pikkuFunc
## Problem

`permissions` on `pikkuFunc` are runtime predicates. Two gaps:

1. **No static surface.** There is no declared vocabulary of "what can be authorized" in an app — nothing to autocomplete, typo-check, render in a console, or grant to a user.
2. **`runPermissions` ORs everything.** `packages/core/src/permissions.ts:329-343` flattens global + wire + tag + function permissions into one list and returns on the *first* pass. A single `addGlobalPermission([signedInUser])` therefore satisfies every function-level permission in the app. Permissions cannot express "this specifically requires admin rights".

## Proposal

A **scope** is a declared, statically-known capability string that must be present on the session:

```ts
wireScope({
  name: 'admin',
  description: 'Administration',
  scopes: { invoices: { scopes: { create: {} } } },
})

export const createInvoice = pikkuFunc({
  scopes: ['admin:invoices:create'],  // tsc-checked against a generated union
  func: async (services, data) => { ... },
})
```

Scopes are an **AND gate that runs before `runPermissions`** — they narrow, never widen, and are never in the permission OR pool. Referencing an undeclared scope is a build error.

## Architecture

- `wireScope` (code) → inspector → generated `ScopeId` union + `SCOPES_META`
- `session.scopes` carries the grants; **core reads, never fetches**
- `ScopeService` resolves scopes at the **session boundary** (better-auth `mapSession`), never in `function-runner`. `mapSession` already runs per request ⇒ realtime, nothing cached, nothing to invalidate.
- Roles are **data** (`pikku_roles`), composed by admins in the console
- `KyselyScopeService` self-creates its tables via `ifNotExists()` (precedent: `KyselySessionStore`), so pikku ships no migration

## Scope of this issue (core only)

- `wireScope` no-op declaration + types
- `verifyScopes` + prefix tree (wildcards: `admin:*`)
- `MissingScopeError` (403)
- `scopes?: string[]` on `CoreUserSession` and `CorePikkuFunctionConfig`
- The gate in `function-runner.ts`, placed after the readonly check and **before** `data()` is awaited — scopes need only the session, so a denied request never parses its body
- `ScopeService` interface

Inspector/CLI codegen, the Kysely service, addon-console functions, and the console UI follow as separate issues.

yasserf:
Closes #956.

Declared, statically-checked authorization scopes on `pikkuFunc`:

```ts
wireScope({ name: 'admin', scopes: { invoices: { scopes: { create: {} } } } })

pikkuFunc({ scopes: ['admin:invoices:create'], func: ... })
//                   ^ autocompleted, and a typo fails the build
```

## Why not just permissions?

`runPermissions` flattens global + wire + tag + function permissions into one
list and returns on the **first pass**. A single `addGlobalPermission([signedInUser])`
therefore satisfies every function-level permission in the app — permissions
cannot express "this specifically requires admin rights".

Scopes are a separate **AND gate**, run before `runPermissions` and before
`await data()` (a denied request never pays to parse its body). Adding a scope
can only ever narrow. A test pins this: a function with a *passing* global
permission must still throw `MissingScopeError`.

## How it fits together

```
wireScope (code)  ──inspector──▶  ScopeId union + SCOPES (generated)
                                        │
                                        ├──▶ pikkuFunc({ scopes: [...] })  tsc-checked
                                        └──▶ pikku_scopes ◀──FK── pikku_role_scopes ──▶ pikku_roles
                                                                                            ▲ FK
                                                                                    pikku_user_role ──FK──▶ user
ScopeService.resolveScopes(userId) ──called by──▶ better-auth session boundary
                                                          ▼
                                                  session.scopes ──▶ verifyScopes ──▶ 403
```

**Core never fetches.** Scopes resolve at the session boundary and the runner
just reads `session.scopes`, so it stays I/O-free (Workers/Lambda). The session
middleware already runs per request, so a grant change lands on the **next
request with no re-login** — nothing cached, nothing to invalidate.

An explicit `scopes` from `mapSession`/`mapKey` is authoritative and never
widened — that's what lets an API key be minted narrower than its owning user.
Impersonation resolves the *target's* scopes, not the admin's.

## Sync is additive

A scope removed from code is **marked, never deleted**. Auto-cascade was
considered and rejected: a rename or rollback would destroy grants permanently,
and during a rolling deploy the last replica to boot would win fleet-wide.
Django solves the identical problem the same way (`remove_stale_contenttypes`
is separate and opt-in). Undeclared scopes are inert — no function can require
one — and `pikku scopes audit` / `pikku scopes prune --yes` are the deliberate
cleanup path, blast radius shown first.

The FK from `pikku_role_scopes` into `pikku_scopes` means the **database itself**
refuses to grant a scope that was never declared.

## Addons

An addon's `wireScope` declarations now merge into the host's `ScopeId` union
and declared set. This required fixing a real shipping bug: tsc only emits a
`.json` into the build output when something imports it, and an addon publishes
only that output — so `pikku-scopes.gen.ts` now imports its metadata sidecar and
derives `SCOPES` from it. The console uses this to be self-hosting: it declares
and requires `pikku:scopes:read` / `pikku:scopes:manage`, so reaching the console
is not the same as being able to grant yourself anything.

## Verification

TDD throughout — every change has a test that failed before it and passes after.

- `verifiers/scopes` — 4 `@ts-expect-error` assertions prove an undeclared scope
  genuinely fails to compile; the inspector independently rejects one even behind
  an `as any`.
- Ordering, fail-fast, and fail-closed gates in core.
- Kysely: additive sync never deletes, FK rejects an undeclared grant, user
  delete cascades, prune revokes.
- CLI: the `--yes` gate is driven through the real command functions against a
  real sqlite database, so it's proven by what lands on disk.
- Suites green: core 1136, inspector 350, cli 349, kysely 134, better-auth 60.

## Follow-ups (not in this PR)

- Console UI — scopes tab + role assignment in `UsersTable`.
- **Pre-existing bug found:** addon `secrets`/`variables` meta sidecars have the
  same shipping problem and never reach a host, making those loader blocks dead
  code for published addons. Deserves its own issue + PR.

## Summary by CodeRabbit

* **New Features**
  * Added statically declared, wildcard-capable scopes with fail-closed enforcement (403 `MissingScopeError`) before request body evaluation and scope-aware TypeScript narrowing.
  * Delivered role- and direct user-scope administration via the console “Scopes” page and user drawers.
  * Added session scope resolution for user, impersonation, and API-key requests, plus add-on scope propagation to hosts.
  * Introduced `pikku scopes audit` and `pikku scopes prune` for stale-scope auditing and cleanup.
* **Bug Fixes**
  * Improved inspector/codegen extraction for `as`/`satisfies` and type-cast arrays so metadata is no longer dropped; secrets/variables scope metadata packaging is now shipped correctly.


yasserf:
Thanks for the review — went through all 14. Summary of what changed in `aefd808d6`:

**Fixed (each with a test that fails before the change):**
- **`loadDeclaredScopes` metadata validation** — now validates the `ScopeDefinitionsMeta` shape and returns `null` on anything else, so a parsed `[]` can no longer become an empty declared set that a confirmed prune would use to wipe every scope.
- **`openScopeService` DB leak** — setup is wrapped; the connection is destroyed before rethrowing if `init()`/`syncScopes()` fails.
- **prune blast-radius count** — counts distinct affected roles (`new Set(stale.flatMap(r => r.roles))`) instead of stale scopes, and the removed-count now reflects rows actually deleted.
- **`pruneScopes` TOCTOU** — the delete now guards on `declared = false` and uses `RETURNING`, so a concurrent `syncScopes` redeclaring a scope between select and delete can't get it (or its grants) deleted.
- **API-key scope-store outage** — scope resolution moved outside the key-verification `try`, so a scope-store failure surfaces instead of silently serving the request anonymously. Now consistent with the human and stateless paths.

**Declined, with reasoning:**
- **Inline comments** — the flagged production comments explain non-obvious *why* (e.g. why `verifyScopes` runs outside `runPermissions`, why the two bootstrap files must precede function types), which the repo guideline explicitly permits, or match an existing convention in the same file (the `// Scopes (...)` label mirrors the neighbouring `// Credentials (...)`/`// Secrets` labels). Removing only these would make them inconsistent.
- **`load-addon-functions-meta` host-wins ordering (#198)** — already reworded; the current comment doesn't claim host-wins. Name collisions are resolved downstream by `validateAndBuildScopeDefinitionsMeta` (dedup, or hard error on differing trees).
- **`load-addon-functions-meta` malformed-catch (#203)** — kept consistent with the sibling secrets/variables blocks in the same file, which use the same bare catch over machine-generated sidecars. Diverging one of three identical blocks isn't worth it.

Issue #966, opened 2026-07-18
Critical security fixes from core audit: SSRF, AI thread/run IDOR, forgeable approvals, channel middleware cross-run leak, workflow step-claim atomicity


yasserf:
A full-surface audit of \`@pikku/core\` surfaced 5 critical findings. This issue tracks fixing them.

**C1 — SSRF in the outgoing-webhook worker** (\`services/queue-webhook-service.ts\`). \`fetch(url)\` on a caller-supplied URL with no private-range/scheme validation and \`redirect: 'follow'\`. Same redirect-bypass gap in \`ai-agent/voice-input.ts\`.

**C2 — AI thread & run IDOR** (\`ai-agent/*\`). \`threadId\`/\`resourceId\`/\`runId\` come from client input and are only checked for existence, never ownership.

**C3 — Forgeable AI approval/credential control-signals** (\`ai-agent/ai-agent-stream.ts\`). \`__approvalRequired\`/\`__credentialRequired\` are read from ordinary tool-result values, so attacker-influenced tool output can forge approval flows with a spoofed display.

**C4 — Channel stream-middleware cache leaks across runs** (\`channel/channel-middleware-runner.ts\`). Per-run middleware closures are cached by agent name → a later run reuses an earlier run's closures (cross-tenant transcript leak + unbounded memory).

**C5 — Workflow step-claim not atomic on Postgres/Mongo** (\`workflow/pikku-workflow-service.ts\`). The \"atomic claim\" relies on \`withStepLock\`, a no-op in the Kysely/Mongo/in-memory backends → duplicate execution of side-effecting steps.

Each fix lands test-first (verifier fails before, passes after).

yasserf:
Per-part status against current `main`. C1-C4 fixed; C5 partly.

**C1 SSRF — FIXED.** `packages/core/src/utils/safe-fetch.ts` (`isPrivateHost`/`assertFetchableUrl`/`safeFetch`, `redirect: 'manual'` with every hop re-validated, Authorization/Cookie stripped cross-origin). Wired at `packages/core/src/services/queue-webhook-service.ts:3,113` and in `voice-input`. Commits `3f7fc543a`, `7ab528738`, `4f62b471f`.

[...]

PR #967, opened 2026-07-18
fix(core): critical security fixes from core audit — SSRF, AI IDOR, approval forgery, channel middleware leak (#966)

yasserf:
Fixes the four **critical** findings from the `@pikku/core` security audit (#966). Each fix ships with a failing-first verifier (test fails before the change, passes after).

## C1 — SSRF in outgoing webhook + voice-input fetch
`safeFetch` (new `packages/core/src/utils/safe-fetch.ts`) blocks requests to loopback / private / link-local / cloud-metadata hosts and **re-validates every redirect hop** (`redirect: 'manual'`), so a public URL can't 302 into `169.254.169.254`. Wired into the webhook delivery worker and the AI voice-input audio fetch, both with an optional `allowedHosts` allowlist (`WebhookServiceConfig.allowedHosts`).

## C2 — AI thread/run IDOR
Thread and run access is now bound to the session owner. `resolveOwnerResourceId` prefers the authenticated `session.userId` over the client-supplied `resourceId`; `assertResourceOwner` throws `ForbiddenError` when a stored thread/run belongs to another owner. Applied across `runAIAgent`, `streamAIAgent`, and both resume paths.

## C3 — Forgeable approval markers
The `__approvalRequired` suspension marker is now only honored from framework sub-agent delegating tools (new `AIAgentToolDef.forwardsApproval` flag). An ordinary tool result — which an attacker-influenced tool/document response can shape — can no longer forge an approval request and suspend the run.

## C4 — Channel middleware cross-run leak
`combineChannelMiddleware` now caches **only** the statically-resolved inherited middleware; per-run `wireChannelMiddleware` is appended fresh each call and never cached, so one request's per-run middleware can't leak into another's.

## Deferred
- **C5** (cross-package DB atomic compare-and-set for workflow step-claim in `@pikku/kysely` / `@pikku/mongo`) is intentionally deferred to its own follow-up PR — it spans other packages.

## Testing
- Full `@pikku/core` suite green (1182 pass / 0 fail).
- `yarn tsc` clean.
- New verifiers: `safe-fetch.test.ts`, `ai-agent-prepare.test.ts` (C2), `ai-agent-runner.test.ts` (C2), `channel-middleware-runner.test.ts` (C4), `ai-agent-stream.test.ts` (C3).

## Summary by CodeRabbit

* **Security**
  * Hardened webhook delivery and voice-audio fetching with HTTP(S) validation, private/internal host blocking, redirect-hop rechecking, and configurable allowed-host allowlists.
  * Prevented forged approval/suspension markers from ordinary tool outputs; approval forwarding now only applies when explicitly enabled for a tool.
  * Enforced AI agent thread/run ownership using session-bound “user” or “org” scoping, including during resume/stream flows.
* **Bug Fixes**
  * Fixed channel middleware caching so per-run middleware closures are not reused across separate runs.

### DIFF
--- a/.changeset/966-c1-c3-hardening.md
+++ b/.changeset/966-c1-c3-hardening.md
@@ -1,0 +1,8 @@
+---
+'@pikku/core': patch
+---
+
+Security hardening (follow-up to the #966 C1/C3 fixes):
+
+- SSRF (C1): `isPrivateHost` now also rejects alias/encoded forms that resolve to internal targets — trailing-dot FQDNs (`localhost.`), the reserved `*.localhost` name, IPv4-mapped IPv6 (`::ffff:127.0.0.1`), octal/decimal/hex-encoded IPv4 (`0177.0.0.1`, `2130706433`, `0x7f000001`), and the full `fe80::/10` link-local range. `safeFetch` also strips `Authorization` and `Cookie` headers whenever a redirect crosses origin so credentials cannot leak to a redirected host.
+- Forgeable approval markers (C3): the sub-agent approval marker is now identified by a non-forgeable `Symbol` brand set only by framework code, instead of the plain `__approvalRequired` string key. A delegating tool's LLM-shaped `result.object` (plain JSON) can no longer conjure an approval/suspension even though the tool is allowed to forward approvals.
--- a/.changeset/966-c2-session-scope.md
+++ b/.changeset/966-c2-session-scope.md
@@ -1,0 +1,5 @@
+---
+'@pikku/core': patch
+---
+
+Security + feature: bind AI agent thread/run ownership to the authenticated session and add `sessionScope: 'user' | 'org'` to agents. The owner key is now the trusted principal (`session.userId`, or `session.orgId` for org-scoped agents) composed with the client `resourceId` (`principal:resourceId`), so a client-supplied `resourceId` sub-partitions within the caller's own boundary but can never widen access to another user's or org's threads. Resolution is idempotent (safe for sub-agent recursion and resume); org scope with no session org is denied; sessionless `user` wirings fall back to the bare `resourceId`.
--- a/.changeset/966-c3-approval-forgery.md
+++ b/.changeset/966-c3-approval-forgery.md
@@ -1,0 +1,5 @@
+---
+'@pikku/core': patch
+---
+
+Security: only honor the `__approvalRequired` suspension marker from framework sub-agent tools (`forwardsApproval`), so an attacker-influenced ordinary tool result can no longer forge an approval/suspension.
--- a/.changeset/966-core-critical-security.md
+++ b/.changeset/966-core-critical-security.md
@@ -1,0 +1,5 @@
+---
+'@pikku/core': patch
+---
+
+Security: SSRF-harden outgoing webhook delivery and voice-input audio fetch (validate scheme + block private/internal hosts, re-validate every redirect hop via a shared `safeFetch`); stop the channel stream-middleware cache from reusing an earlier run's per-invocation middleware closures across runs.
--- a/packages/core/src/services/queue-webhook-service.test.ts
+++ b/packages/core/src/services/queue-webhook-service.test.ts
@@ -241,6 +241,21 @@ describe('pikkuWebhookWorkerFunc', () => {
     )
   })
 
+  test('refuses to deliver to a private/internal host without fetching it (SSRF)', async (t) => {
+    t.after(restoreFetch)
+    const requests = stubFetch(() => ({ status: 200 }))
+
+    await assert.rejects(
+      pikkuWebhookWorkerFunc({} as any, {
+        url: 'http://169.254.169.254/latest/meta-data/',
+        body: '{}',
+        headers: {},
+      }),
+      /private\/internal host/
+    )
+    assert.equal(requests.length, 0, 'must never issue the request')
+  })
+
   test('records a delivered attempt via the store when a deliveryId is present', async (t) => {
     t.after(restoreFetch)
     stubFetch(() => ({ status: 200 }))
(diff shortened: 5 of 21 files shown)


yasserf:
### CI note

The first CI run surfaced three failures; two were mine and are fixed in `2a544feb8`:

- **Verifiers (webhooks)** — the new C1 SSRF guard correctly blocked the test's `127.0.0.1` mock receiver. Fixed by opting the verifier into the `config.webhook.allowedHosts` allowlist (the intended escape hatch for internal delivery).
- **Verifiers (middleware-and-permissions)** — the C2 owner check caught an internally-inconsistent test mock (`MockAIStorage.getThread` returned `resourceId: 'test'` while the caller owned `'test-resource'`). Aligned the mock's resourceId. Both verifiers pass locally.

**E2E AI** has one failing scenario (`todo-agent.feature:47` "Delete a todo shows approval with reason" / "Buy groceries"). This is **pre-existing on `main`, not from this PR** — it landed with #925 (outgoing webhooks), whose own E2E AI run failed byte-for-byte identically ("24 scenarios (1 failed, 23 passed)", same 4 `No pending approval on sub-agent run` logs) and was merged red. The failing path is a direct `needsApproval` tool approval that this PR's C3 gate never touches. Tracking it for a separate follow-up.

yasserf:
@coderabbitai review

yasserf:
### C2 reworked: `sessionScope` + composite ownership (commit 8db6d0296)

Replaced the blunt C2 override (`resourceId → session.userId`, which ignored the client value and would have broken shared/org-scoped threads like Fabric's org-observer) with a per-agent, Mastra-aligned ownership model:

- **`CoreAIAgent.sessionScope: 'user' | 'org'`** (default `'user'`).
- Owner key = **trusted principal composed with the client `resourceId`**: `principal:resourceId`, where principal is `session.userId` (`'user'`) or `session.orgId` (`'org'`). The principal is always the prefix, so a client `resourceId` sub-partitions *within* the caller's boundary (`alice:project-1`) but can never forge another owner (`bob` passing `alice:x` → `bob:alice:x`).
- **Idempotent** resolution — sub-agent recursion and resume pass an already-composite `resourceId`; the trusted-prefix check prevents double-composing (`alice:alice:default`).
- **`'org'` with no session org → `ForbiddenError`** (composing a bare/shared key would leak across orgs). Sessionless `'user'` falls back to the bare `resourceId`.

Dropped `'team'` (no consumer; Better Auth teams is a plugin option that's hard to statically guard) and a free-form resolver `fn` (the composite covers per-user sub-partitioning and stays serializable into meta, so sub-agents/queued runs honor it).

TDD: 9 failing verifiers before, 20 passing after (`ai-agent-prepare.test.ts`). Full core suite green (1190/0), tsc clean.

**Out of scope (separate follow-up):** the thread **list/read** endpoints (`getAgentThreads`/`getAgentThreadMessages`/`getAgentThreadRuns`) currently enforce no owner at all — a pre-existing IDOR orthogonal to this change; owner-scoping them spans the CLI templates + console addon and needs its own design.

yasserf:
### Rebased onto latest `main` · CI board now 89 pass / 1 fail

Rebased `fix/966-core-critical-security` onto current `origin/main` (picks up #960 scopes + #968 console) — **clean rebase, no conflicts** — and force-pushed (`4fe5e5347`).

**Fresh CI run:** [29681827172](https://github.com/pikkujs/pikku/actions/runs/29681827172) → **89 pass, 1 fail**.

- The earlier `Verifiers (types)` red was a **network flake** (`ECONNRESET` fetching the `serverless` binary during `yarn install`) — cleared on rerun. ✅
- `Verifiers (dev-bun)` passes in CI (a local pre-push-hook-only failure from bun runtime auto-detect on the dev box; not a code issue). ✅
- The sole remaining red is **E2E AI**, and it is **byte-identical to the pre-existing failure on `main`**: `todo-agent.feature:47` "Delete a todo shows approval with reason" → `✖ approval reason should contain "Buy groceries"`, "24 scenarios (1 failed, 23 passed)". This landed via #925 (merged red) and is **not a regression from this PR**. `main` has no required status checks, so #967 is `MERGEABLE`.

The E2E AI approval-reason bug is a separate #925-scoped concern (root cause localized: `deleteTodo`'s `approvalDescription` runs against the wrong package `singletonServices`, so `todoStore.get(id)` throws and the reason is silently swallowed at `ai-agent-stream.ts:297-299`). It should be fixed in its own branch/PR, not folded into the security work.

yasserf:
### CodeRabbit hardening on the C1/C3 fixes (commit 59ebcafe2)

Follow-up commit closing the review gaps on the security fixes already in this PR (TDD: red before, green after; core suite 1242 pass / 0 fail, tsc clean):

**C1 SSRF — `isPrivateHost` alias coverage** (`utils/safe-fetch.ts`)
- Now rejects: trailing-dot FQDN (`localhost.`), `*.localhost`, unspecified `::`, IPv4-mapped IPv6 (`::ffff:127.0.0.1` and the `::ffff:7f00:1` form the URL parser normalizes to), octal/decimal/hex-encoded IPv4 (`0177.0.0.1`, `2130706433`, `0x7f000001`), and the full `fe80::/10` link-local range (previously only literal `fe80:`).
- Public look-alikes still allowed (`134744072` = 8.8.8.8, `2001:db8::1`, deprecated `fec0::` site-local).

**C1 — cross-origin credential stripping** (`safeFetch`)
- `Authorization` and `Cookie` are dropped when a redirect crosses origin (preserved same-origin), so credentials never leak to a redirected host.

**C3 — Symbol-branded approval marker** (`ai-agent-prepare.ts` producer, `ai-agent-stream.ts` consumer)
- The sub-agent approval marker is now identified by a non-forgeable `APPROVAL_REQUIRED` Symbol set only by framework code, instead of the `__approvalRequired` string key. A delegating (`forwardsApproval`) tool's LLM-shaped `result.object` is plain JSON and cannot carry a Symbol, so it can no longer conjure an approval/suspension even though the tool is permitted to forward approvals.

The DNS-rebinding limitation (a public hostname that *resolves* to a private IP) remains documented in the `safe-fetch.ts` header — it needs DNS resolution unavailable in edge runtimes.

PR #984, opened 2026-07-20
feat(core): add docsUrl to secret/variable/credential wirings

yasserf:
## Why

A config value can't currently say where you get it. A deploy UI reporting a missing `STRIPE_SECRET_KEY` has only the identifier to show, so the user is left to guess or go ask someone.

## What

Adds one optional field to `wireSecret`, `wireVariable`, and `wireCredential` (and their `*DefinitionMeta` twins):

```ts
wireSecret({
  name: 'stripeKey',
  displayName: 'Stripe Secret Key',
  secretId: 'STRIPE_SECRET_KEY',
  docsUrl: 'https://dashboard.stripe.com/apikeys',
  schema: StripeKeySchema,
})
```

Absent by default, so existing wirings are unaffected.

## The plumbing is the point

`docsUrl` is threaded through the inspector (`add-keyed-wiring`, `add-credential`) **and** the three definition validators, which rebuild the meta object field-by-field and would otherwise silently drop it.

That failure mode isn't hypothetical: `displayName` is declared, inspected, and validated today, and still gets dropped by at least one downstream consumer — so UIs show the raw identifier even though a human label exists. The tests here assert `docsUrl` survives inspection for all three wiring kinds, specifically so it can't rot the same way.

## What this PR deliberately does *not* add

An earlier revision also added an `optional` flag, motivated by a platform-injected secret (`CREDENTIALS_KEY`) that was blocking deploys behind a value no human can know. That was the wrong abstraction and has been removed.

`optional` asserts *the app runs without this value*. For a platform-injected secret that's simply false — the app requires it; the platform mints it instead of the user. "Someone else provides this" and "you don't need this" are different claims, and conflating them would have baked a misleading flag into every pikku deployment to serve one platform's provisioning detail.

Which entity supplies a value is a fact about a deployment platform, not about the app, so it belongs in that platform.

## Testing

`yarn workspace @pikku/inspector test` — 389 pass, 0 fail (4 new).

Note for reviewers: the new tests write fixtures **inside** the package rather than in `os.tmpdir()`. In a temp dir `zod` doesn't resolve, the schema-vendor check rejects the schema, and the definition is never recorded — so a naive fixture location makes these tests fail for a reason unrelated to what they're testing.

## Note on CI

The Build job failure on this branch is not from this branch — it reproduces on a clean `origin/main`. Fix is in #986.

## Summary by CodeRabbit

* **New Features**
  * Added optional documentation links for secrets, variables, and credentials.
  * Configuration interfaces can now direct users to where missing values can be retrieved.
  * Documentation links are preserved in inspected configuration definitions.

* **Tests**
  * Added coverage confirming documentation links are retained and omitted links remain undefined.


yasserf:
Lint is fixed in 7500a6da1 (type-only import in the new test).

The Build failure here is not from this branch — it reproduces on a clean `origin/main`. Fix is in #986; this PR's build will stay red until that merges.

PR #1002, opened 2026-07-22
feat(better-auth)!: gate admin capabilities on scopes, and scaffold user management

### LINKED ISSUE #1001: better-auth: default impersonation gate should be a scope, not `role === 'admin'`
## Problem

`resolveImpersonatedSession` gates impersonation on a sweeping role by default:

```js
// packages/better-auth/src/auth-session-impersonation.ts
const canImpersonate =
  impersonation.canImpersonate ?? ((result) => result.user?.role === 'admin')
```

`user.role === 'admin'` is a single text column — one blob that means "can do
everything". Impersonation is a *capability*, and pikku already has the right
primitive for capabilities: scopes. `withResolvedScopes` (same package) resolves
`session.scopes` from a registered `ScopeService` at the session boundary, so the
value is already present on the session by the time this gate runs.

The two features exist; they've never been connected.

## Proposal

Make the DEFAULT scope-aware, keeping the role check as the fallback:

- If a `ScopeService` is registered, default `canImpersonate` to
  `session.scopes.includes('auth:impersonate')`.
- If it isn't, keep `user.role === 'admin'` exactly as today.
- `impersonation.canImpersonate` stays overridable and authoritative.

Backwards compatible: an app with no `ScopeService` sees no change, and an app
that passes its own gate is unaffected.

## Why it matters

A role is not a permission. "Who may impersonate" and "who may administer
billing" are different capabilities that a single `role` column cannot express,
and a user can legitimately hold several. Deriving scopes from roles in
`mapSession` keeps roles as a *grant source* while the enforcement point stays a
capability check — which is what the rest of pikku already does via `scopes: []`
on a function.

## Scope

Small: the default in `auth-session-impersonation.ts`, plus a test for each of
the three paths (explicit gate / ScopeService present / no ScopeService).

Note the sibling concerns that are deliberately NOT in scope here: `ban` and
`listUsers` are actions an app *invokes*, not things pikku gates — those belong
on the app's own functions via `scopes: ['users:ban']`, with no upstream change.

### LINKED ISSUE #1006: Migrate the remaining scaffolds to directory-per-domain with zod schemas
Every scaffold except `admin` still emits a single flat `*.gen.ts` that declares its payload shapes as inline TypeScript type literals with generics. They should all move to the directory-per-domain layout PR #1002 introduced, so their schemas are zod.

### Why zod needs its own module

The inspector converts a zod schema by *importing the module that declares it* at codegen time. It cannot import a scaffold's functions file: that file imports pikku-types over a relative path which per-unit deploy codegen rewrites (e.g. into `.deploy/serverless/units/*/.pikku/…`), and the rewritten path goes stale. A sibling module that imports nothing but `zod` sidesteps this entirely — which is why the layout is a directory, not just a rename.

This is what made the earlier "scaffolds use TS generics" decision necessary; #1002 removed the blocker, so the exception can go.

### The layout

`scaffold/admin/` is the reference:

```
src/scaffold/admin/user-admin.gen.ts          # functions, imports the schemas
src/scaffold/admin/user-admin.schemas.gen.ts  # exported zod consts, imports only `zod`
```

Per scaffold that means:
- the serializer returns `{ schemas, functions }` instead of a string (`serializeUserAdminFunctions` is the model; `serializeAuthGen` is the pre-existing multi-file precedent)
- the command writes both files
- `pikku-cli-config.ts` derives both paths under a domain directory
- `types/config.d.ts` gains the `…SchemasFile` sibling

### Remaining scaffolds

| scaffold | current file (`packages/cli/src/utils/pikku-cli-config.ts:318-383`) | suggested directory |
| --- | --- | --- |
| `graph` | `graph.wirings.gen.ts` | `scaffold/graph/` |
| `webhook` | `webhook.gen.ts` | `scaffold/webhook/` |
| `workflow` | `workflow-routes.gen.ts` | `scaffold/workflow/` |
| `rpc` | `rpc-public.gen.ts` | `scaffold/rpc/` |
| `remoteRpc` | `rpc-remote.gen.ts` | `scaffold/rpc/` |
| `agent` | `agent.gen.ts` | `scaffold/agent/` |
| `console` | `console.gen.ts` | `scaffold/console/` |
| `events` | `events.gen.ts` | `scaffold/realtime/` |
| `scenarios` | `scenarios.gen.ts` | `scaffold/scenarios/` |
| auth (always on) | `auth.gen.ts` | `scaffold/auth/` |

Basenames stay as they are — only the nesting and the schemas split change.

### Notes

- Scaffold `func:` must remain an **inline arrow**. A `func:` that references an identifier imported from node_modules leaves the inspector unable to resolve the handler body, silently dropping `sessionless`/`funcWrapper` from meta, which surfaces at runtime as a 403 "Authentication required".
- Zod schemas must be **exported** consts — the converter imports them by export name.
- This moves generated files, so templates and any project with a committed `src/scaffold` will need a regenerate; worth doing in one pass rather than trickling.
- Verification per scaffold: after regenerating, the schema JSON for each function's input/output should appear under `.pikku/schemas/schemas/`. Its absence means the inspector failed to resolve the zod module.

Follow-up to #1001 / #1002.

yasserf:
Closes #1001.

#1001 proposed making the impersonation default scope-aware while keeping `role === 'admin'` as a fallback. This goes further: **the role stops being an authority**. Every admin capability becomes a scope, and better-auth's `user.role` survives only as a *projection* of the scope store.

The role column said one thing — "can do everything". But impersonating a user, rebinding a shared credential, reading the user directory and banning someone are distinct capabilities one person can hold independently, and a text column cannot express that.

| Gate | Scope |
| --- | --- |
| impersonation default | `admin:impersonate` |
| `credentialOAuth`'s `canLinkSingleton` | `admin:credentials:link` |
| reading the user directory | `admin:users:list` |
| creating a user out of band | `admin:users:create` |
| ban / unban | `admin:users:ban` |
| delete a user | `admin:users:remove` |
| revoke a user's sessions | `admin:users:sessions` |
| set a user's password | `admin:users:password` |

Holding the bare `admin` scope satisfies all of them through pikku's existing parent-grant rule, so it is a one-for-one replacement for the old role.

## `admin()` stays — the role becomes a projection

An earlier revision of this PR dropped better-auth's `admin()` plugin outright. That was wrong: it is what *implements* ban, delete, session-revocation and set-password, and those are real capabilities, not role trivia. The plugin is back.

What does not come back is role-as-authority. `user.role` is now derived, never granted:

- It is synced at the session boundary (`syncProjectedAdminRole`), where scopes are already resolved — one code path covers every way a grant can change (direct grant, role membership, `setRoleScopes`, `deleteRole`, prune) with no fan-out to miss and zero writes in the steady state.
- It is projected from **exactly the scopes whose capability better-auth itself gates on the caller's role** — `admin:users:{create,ban,remove,sessions,password}`. Deliberately *not* from `admin:impersonate` or `admin:users:list`: the directory read goes straight to the auth adapter and needs no role, and neither reading nor impersonating should silently confer the power to ban or delete. Root `admin` still reaches them via parent-grant.
- Only the human sign-in path projects it. An API key may carry a deliberately narrowed scope set, and an impersonated session acts *as* the target rather than on their behalf — neither should rewrite what a user may do when signed in themselves.
- Revoking the scope demotes the role on the next sign-in.

`null` is not reachable, so a non-admin is better-auth's own `'user'` default: the plugin's create hook does `role: options?.defaultRole ?? "user"` and its permission check does `(input.role || ... || "user")`, so an empty string coerces straight back. The projection auto-detects the plugin from the resolved instance rather than taking a config flag — a flag can be forgotten, and forgetting it ships exactly the drift the projection exists to prevent.

## User management is scaffolded, not an addon

The whole set lives in **your app** (`scaffold.userAdmin` in `pikku.config.json`), not in `@pikku/addon-console`:

```
pikkuAdminListUsers      pikkuAdminCreateUser
pikkuAdminSetUserBanned  pikkuAdminRemoveUser
pikkuAdminRevokeUserSessions  pikkuAdminSetUserPassword
```

Listing or banning a user is ordinary application behaviour. Nobody should have to install a console — which exists to be served alongside a running dev server — just to get it. Same motivation as `serialize-scenario-functions.ts`, which already works this way.

The scaffold is a **directory** — `scaffold/admin/user-admin.gen.ts` beside `user-admin.schemas.gen.ts` — and the functions take zod `input`/`output` schemas rather than TS generics. The schemas have to stand alone in a module that imports nothing but zod: the inspector reads a zod schema by importing the module that declares it, which it cannot do for a file whose relative pikku-types import per-unit deploy codegen rewrites.

Each is gated on its own `admin:users:*` scope, brokers through `callAdminApi` in `@pikku/better-auth` (forwarding the caller's headers, because `admin()`'s routes authorize the *caller's* role via `adminMiddleware`), and codegen **fails with an actionable error** if better-auth is wired without `admin()` — otherwise the functions would throw on every call and the four scopes would grant nothing.

The console's Users page calls these same functions when they exist, showing each action only where the caller holds its scope, plus a ban/status column that appears only where the host actually wires `admin()`.

## Every scaffold moves to that layout (closes #1006)

This started as a follow-up and is now done here, because leaving nine scaffolds on TS generics would have made the exception look like the rule. `graph`, `webhook`, `workflow`, `rpc`, `remoteRpc`, `agent`, `console`, `events`, `scenarios` and `auth` each emit a directory named for their domain, wiring file beside a `*.schemas.gen.ts` sibling.

Three things fell out of it:

- **The agent input duplication is fixed.** `agentCaller` and `agentStreamCaller` take the same payload but had to repeat the type literal verbatim in each generic position: the extractor synthesised the schema name from the *function* name, so behind a shared named alias it recorded an `inputSchemaName` with no schema behind it and every call through that alias failed at runtime with `MissingSchemaError`. A zod `input` resolves by reference — one `AgentCall` schema now backs both.
- **Core-typed results carry no `output` schema.** Where the shape belongs to `@pikku/core` — `WorkflowRunStatus`, `FunctionCoverageReport`, `StubCall[]` — the generated function omits `output` and the inspector infers it from the handler's return type. Re-declaring a core type in zod would be a second definition free to drift; the generated http-map still resolves `WorkflowRunStatus` by name.
- **Upgrading is a single codegen run.** The old flat file still wires the same routes, so a leftover copy wires everything twice and the run dies on PKU851. The prune therefore happens *before* the first inspection, not after each write — a cleanup that ran after the write left the first run red and only the second one green, which is exactly the failure this was caught on.

## Why the impersonation gate resolves scopes instead of reading `session.scopes`

The issue assumed scopes are already on the session when the gate runs. They are not: `resolveImpersonatedSession` runs against better-auth's raw `{ user, session }`, *before* `mapSession`/`withResolvedScopes`. So the gate resolves the caller's grants itself, via `resolvedUserHoldsScopes(userId, required, scopeService, logger)`.

It fails closed. With no `ScopeService` registered nothing can be granted, so nothing is authorized — and that denial is logged at `warn`, because an app wiring an admin capability with no `ScopeService` has a configuration bug, not a permissions problem.

## The directory is brokered rather than proxied — and it is not the console's

`pikkuAdminListUsers` reads through better-auth's DB-agnostic `$context.adapter` instead of `client.admin.listUsers`, so it needs no admin API surface, works on any database better-auth supports, and carries ban state when the plugin is present.

It is scaffolded, not shipped: `@pikku/addon-console` no longer has a `console:listUsers` function at all. Reading your own user directory is not something an app should install a console to do. The console's Users page consumes the scaffolded RPC like any other client.

`console:getMyAccess` returns the caller's resolved scopes so the UI can ask what it *may do* rather than what role it *has* — `can('admin:impersonate')` honours the parent-grant rule.

## Two codegen fixes that fell out of this

- **`flattenScopeDefinitions` now dedupes by id.** A root declared by both an addon and its host was emitted once per declaration. The validator correctly accepts identical redeclarations, but the serializer wrote them into an object literal — a `TS1117` duplicate-key error, plus an inflated `SCOPES` sync and `ScopeId` union.
- **`auth` is declared on `CoreSingletonServices`.** The generated `pikkuServices` wrapper already injected it (and already forbade service factories from returning it, via `Omit<RequiredSingletonServices, 'auth'>`), but nothing typed it — so a function that destructured it did not compile.

## Breaking changes

- **No role fallback for the scope-gated capabilities.** An app relying on the old default must register a `ScopeService` and grant `admin` (or a narrower `admin:*`).
- `delegatedAuth`'s `defaultRole`/`mapRole` keep their names and semantics but grant a *pikku* role through the `ScopeService` instead of writing better-auth's `role` column.
- The `credentialOAuth` platform user no longer sets `banned`. It was already unreachable for a better reason: it is created with no credential account at all.
- `UsersTableUser` / `UsersTableLabels` lose `role` and gain a required `emptyCreated`. **This breaks Fabric's stage Users tab**, which lives outside this repo — follow-up needed there once this releases.

Every grant path (dev quick-login, fabric, delegated auth) warns rather than throws when it cannot grant: a provisioning gap degrades to an authenticated-but-unprivileged session, which is a far clearer symptom than a 500 on sign-in.

## Coverage

New `@user-admin` RPC suite (7 scenarios, 33 steps): a caller without the scope is refused for each of the four actions; a scoped caller bans a user, sign-in is then blocked, and lifting it restores access; sessions are revoked; and holding **only** `admin:users:list` does *not* confer ban.

That last one plus a mutation test is what makes this more than a smoke test — forcing `projectedAdminRole` to always return the default role turns both success scenarios red (500 from better-auth's own gate), so the suite proves the projection reaches the plugin rather than better-auth simply not caring.

`@impersonation` covers the three paths that gate must get right:

1. a holder of `admin:impersonate` runs as the target;
2. a caller without it who forges the same header is silently ignored and keeps running as themselves — a forged header can never widen access;
3. granting the scope directly opens the gate on the next request and revoking it closes it, which is what proves the **scope**, not the identity, is the gate.

New `@user-admin-console` browser suite drives the whole lifecycle through the real UI — create → ban → unban → revoke sessions → set password → delete — asserting the directory and the status column after each step and that the target's ability to sign in tracks the ban. It deliberately carries *no* partial-scope scenario: the console gates entry on the umbrella `admin` scope, which parent-grants every leaf, so a partially-scoped caller never reaches the page to be observed. That separation is covered where it is observable, over RPC.

`console-authz.feature` gains the `admin:users:list` cases. Unit coverage: `admin-role-sync.test.ts` (11, including the mutation-sensitive "impersonate alone does not promote" and "list alone does not promote"), `user-admin-functions.test.ts` (9, including the missing-`admin()` guard), and two new `flattenScopeDefinitions` dedupe cases.

The console's user-management surfaces are drawers, not modals, matching the rest of the console.

Verified green: 87/87 default e2e (465 steps), `@user-admin`+`@user-admin-console`+`@console-authz`+`@scopes` 18/18 (97 steps), `@agent-protocol` 55/55 (178 steps), cli 429, repo-wide `yarn build` and `yarn test:verifiers`. The scaffold move was re-verified against three real projects — e2e, `templates/functions` (which additionally exercises `realtime/` and `rpc/rpc-remote`) and `verifiers/better-auth` — each regenerating clean in a single run from the pre-move layout.

## Summary by CodeRabbit

* **New Features**
  * Switched console and admin authorization to scope-based checks for impersonation, credential linking, and user management.
  * Added console access inspection (e.g., new “who am I” style identity endpoints) and “get my access” for returning current scopes.
  * Enabled scaffolded user-administration endpoints and expanded console user-management UI actions (create, ban/unban, sign out everywhere, reset password, delete).

* **Bug Fixes**
  * Prevented forged impersonation headers and unauthorized admin actions; authorization now fails closed when scope resolution isn’t available.

* **Documentation**
  * Updated admin/security/permission guidance to reflect “admin scopes, not roles.”


yasserf:
@coderabbitai review

yasserf:
Thanks — worked through all 8. Pushed in 79275e3be.

**Taken (7):**

1. **`hooks.ts` readiness probe** — right, `status !== 403` treats a 500 during startup as ready. Now `ping.ok`. Verified live: `@impersonation or @console-authz or @scopes` → 14 scenarios, 14 passed.
2. **`wire.session.get()`** — correct, and worse than flagged. `PikkuWire` exposes `session` as a *value* plus `setSession`/`clearSession`, so `session.get()`, `session.set()` and `session.clear()` were all wrong. Fixed all five sites across three skill files, not just the one in the diff — these docs are what agents read as the source of truth for the API.
3. **`pikku-security` `setSession`** — agreed, and sharper than "leftover guidance": writing `role: user.role` into the session is the exact pattern this PR argues against. Dropped from both examples.
4. **Sidebar `asI18n` literals** — genuine violation of the repo's own rule. Added `nav_scenarios` / `nav_webhooks` / `nav_scopes` message keys.
5. **`UsersTable` `asI18n('—')`** — moved into `UsersTableLabels` as `emptyCreated` rather than rendering bare, since the component's contract is that the host supplies all translated copy.
6. **Dead `docsHref`** — confirmed 404, along with `authentication/permissions` and `authentication` itself. Repointed at `core-features/permission-guards` (verified 200), and fixed the identical dead link already sitting in `ScopesPage.tsx`.
7. **Fixture-restating comments** — removed both.

**Not taken (1):**

**Seeded emails in the seed logs.** These are static test fixtures — `admin@e2e.test`, `staff@e2e.test`, `guest@e2e.test` — hardcoded in `auth-fixtures.ts`, not user data, and the repo is public. Naming which fixture received which grant *is* the diagnostic value of the line; replacing it with a count would make the log useless for exactly the seeding-order bug the readiness probe above exists to catch. The CWE-532 match is a true pattern match on a false positive.

PR #1004, opened 2026-07-22
fix(cli): remove generated schema files that are no longer required

yasserf:
## The bug

`saveSchemas` rewrites `register.gen.ts` from scratch on every run, so a schema that is no longer required stops being registered — but its `<Name>.schema.json` stayed on disk indefinitely.

That orphan is not inert. It is the artifact tooling reaches for to answer *"what does the server validate against?"*, and it answers with the shape the function had at some earlier point.

## Why it matters

A consumer trusting the file concludes the schema is correct while the running server rejects the very field the file lists. That presents as an unfixable "stale server":

- regenerating does not remove the file, so the contradiction survives
- restarting does not change what the file says, so the advice it implies never works

We hit this in an automated build agent, which spent its entire budget on one rejected call — repeatedly restarting and rewriting a function whose on-disk schema looked correct the whole time, because the orphaned file was the thing being read.

## The fix

Keep the schemas directory in step with `register.gen.ts`: if a schema is not imported there, its file is deleted. This also covers the case where nothing is required any more, where every file in the directory is an orphan (`register.gen.ts` is already reduced to `export const empty = null`, so leaving the JSON behind is exactly the misleading state).

Failures to unlink are logged rather than thrown — stale output should not fail a build, but it must not be silent either.

## Tests

`packages/cli/src/utils/serialize-schemas.test.ts`, 5 cases:

| test | asserts |
|---|---|
| removes schema files that are no longer required | orphans are deleted |
| the surviving file is this run's output | not the stale copy |
| every schema file is registered in `register.gen.ts` | the disk/register invariant |
| clears every schema file when nothing is required | the empty case |
| leaves an unrelated file alone | only `*.schema.json` is touched |

**3 of the 5 fail without the fix**; the other 2 are controls that should pass either way.

All 54 tests in `src/utils/` pass. Other suites in the package fail on a clean worktree for an unrelated pre-existing reason (they import `dist/.pikku/*.gen.js`, which needs a build).

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed stale generated schema files when schemas are deleted or no longer registered.
  * Kept schema files synchronized with generated registrations to prevent outdated validation artifacts.
  * Preserved unrelated files in the schemas directory during cleanup.

### DIFF
--- a/.changeset/olive-cups-shave.md
+++ b/.changeset/olive-cups-shave.md
@@ -1,0 +1,16 @@
+---
+'@pikku/cli': patch
+---
+
+Remove generated schema files that are no longer required
+
+`saveSchemas` rewrites `register.gen.ts` from scratch on every run, so a schema that
+is no longer required stops being registered — but its `<Name>.schema.json` stayed on
+disk indefinitely. That orphan is not inert: it is the artifact tooling reads to answer
+"what does the server validate against?", and it answers with the shape the function had
+at some earlier point. Anything trusting it concludes the schema is correct while the
+running server disagrees, which presents as an unfixable "stale server" that no
+regeneration or restart resolves.
+
+The schemas directory is now kept in step with `register.gen.ts`: if a schema is not
+imported there, its file is deleted.
--- a/packages/cli/src/functions/wirings/functions/schemas.ts
+++ b/packages/cli/src/functions/wirings/functions/schemas.ts
@@ -6,7 +6,18 @@ export const pikkuSchemas = pikkuSessionlessFunc<void, boolean | undefined>({
   func: async ({ logger, config, getInspectorState }) => {
     const visitState = await getInspectorState()
 
-    if (Object.keys(visitState.schemas).length === 0) {
+    // Bodies missing while functions still DECLARE schemas means the inspection pass
+    // was partial. Writing from it would empty register.gen.ts and delete the files
+    // behind schemas that are still required — unregistering the whole app off a bad
+    // read. Leave the previous output alone.
+    //
+    // Zero REQUIRED schemas is a different thing entirely and must still be written:
+    // that is a project whose last schema was removed, and it is the case where the
+    // files left by earlier runs have to be cleared.
+    if (
+      visitState.requiredSchemas.size > 0 &&
+      Object.keys(visitState.schemas).length === 0
+    ) {
       return undefined
     }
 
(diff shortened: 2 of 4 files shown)


yasserf:
Thanks — went through all four. Three were right and are fixed in `ebd48416b`; one I fixed differently than suggested, and I want to flag why.

**1. `readdir` catch too broad** — fixed. ENOENT returns (first run / never generated); anything else throws. A permissions or IO error meant we hadn't read the directory at all, so reporting a clean generation over it was exactly the unearned claim this PR exists to stop.

**2. Own-key presence over truthiness** — fixed, using `schemas[schema] !== undefined`. `false` is a valid JSON Schema and the old filter both unregistered it and pruned its file. I preferred `!== undefined` to `hasOwnProperty` because an explicitly-`undefined` value would survive the latter and then `JSON.stringify` to `undefined` in the write loop. New test `keeps a \`false\` schema` fails without the change.

**3. Empty-schema cleanup unreachable** — confirmed and fixed, but *not* by calling `saveSchemas` unconditionally. That early return is load-bearing: `saveSchemas` keys its empty branch on `requiredSchemas` while the caller guarded on `schemas`, and the state where those disagree — no bodies generated while functions still *declare* schemas — is a partial inspection pass. Always calling would write an empty `register.gen.ts` and delete every file, unregistering the whole app off a bad read.

So the guard now keys on the condition it was actually reaching for:

```ts
if (visitState.requiredSchemas.size > 0 && Object.keys(visitState.schemas).length === 0) return undefined
```

Zero *required* schemas is a genuinely empty project and reaches the prune; the partial-pass case still leaves the previous output alone.

**4. Unlink-failure contract** — added. `reports a schema it could not delete instead of failing the build` seeds a directory named `*.schema.json` (unlink can't remove it) and asserts `saveSchemas` resolves *and* names the file via `logger.error`. A stale file we can't remove is not a build failure, but it must not be silent.

**Nitpick** — dropped the two behaviour-restating comments.

Not done: a wiring-level test for `pikkuSchemas` itself. It needs the generated `#pikku` types (`dist/.pikku/pikku-types.gen.js`), which aren't present in a source-only checkout, so every existing test in that directory fails the same way — I'd rather not add an eighth. The unit tests cover both `saveSchemas` paths; the caller change is the one-line guard above.

7/7 in `src/utils/serialize-schemas.test.ts`; `tsc --noEmit` clean on the changed files (the `#pikku` TS2307 in `schemas.ts` is pre-existing and build-generated).

PR #1009, opened 2026-07-22
feat(console): scroll the addon and API catalogues instead of loading them whole

yasserf:
## Why

The APIs tab fetched a fixed first 100 entries out of ~2,500 and never fetched more, so most of the catalogue was unreachable and its search box only ever searched those 100. The addons tab loaded all 214 entries up front — 21 MiB of JSON, because each row carried its entire stored registry entry.

## What

Both galleries now use TanStack `useInfiniteQuery` and pull the next page as the grid scrolls (IntersectionObserver sentinel, 50 rows/page).

Because a paged list can only be filtered honestly by the server, search, category, sort and the All/Official/Installed filter all moved to the registry. The category rail's counts now come from a catalogue-wide facet call rather than being derived from whatever rows happened to be loaded.

`CommunityGallery` is now fully controlled — all client-side filtering and sorting is gone.

## Breaking (`@pikku/addon-console`)

`getAddonMeta` took no input and returned `AddonMeta[]`. It now takes `{ cursor?, limit?, search?, category?, sort?, official?, names? }` and returns `{ packages, total, nextCursor }`. Callers wanting the whole catalogue should walk `nextCursor` — the `useAddonMeta` hook does this and still returns a flat array, so `CanvasDrawerFactory`'s call sites are unchanged.

Adds `getAddonCategories` and `getOpenapiCategories`, and `category` to `getOpenapis`.

## Deploy coupling

**This needs the matching fabric registry change deployed first** (pikkufabric/fabric#…, same branch name). Against an older registry the new query params are rejected with 422. The second commit here makes the one degradable case degrade — a registry without `/registry/packages/categories` falls through to `/registry/packages/:id`, which 204s with an empty body, and `.json()` was throwing on it.

## Notes

- Category counts now count *every* declared category, not just `categories[0]`. This makes the counts agree with the filter (which always matched any category), but the numbers are higher than before.
- Payload: full catalogue 21.12 MiB → 335 KiB; per 50-row page 1421 KiB → 54 KiB; worst single row 3.82 MB → 27 KB. Every field the grid reads is retained.

## Summary by CodeRabbit

* **New Features**
  * Addon and API galleries now incrementally load more results as you scroll (infinite pagination).
  * Search, category, sorting, and Official/Installed filters now return accurate catalogue-wide results.
  * Category counts and total available results are displayed consistently.
  * API/OpenAPI listings now support category filtering.
* **Bug Fixes**
  * Improved empty states differentiate unavailable catalogues from searches with no matches.
  * More reliable handling when category responses are empty.


yasserf:
@coderabbitai review

PR #1018, opened 2026-07-22
fix(cli): validate fixtures construct a Kysely so the db-adapter check fires

yasserf:
## What

`c42d7ec3` narrowed the `services-wrong-db-adapter` validator so it only flags a `services.ts` that actually **constructs** its own client (`new Kysely(`) — the starter scaffold takes kysely from injection and constructs nothing, and flagging it told every fresh project to import a dialect it must not use. Correct change.

But the two db-adapter test fixtures only *imported* `Kysely`, so after that narrowing the check no longer fired against them:

- `uses Kysely without LibsqlWebDialect → error` started **failing** (finding no longer emitted)
- `uses LibsqlWebDialect → no adapter error` started passing **vacuously** (it would pass whether or not the guard worked)

This is what turns `main` red on the CLI test suite right now.

## Fix

Give both fixtures a real `new Kysely({ dialect: ... })` so they exercise the guard as written — the postgres one trips the finding, the libsql one constructs the correct dialect and stays clean.

## Verification

`node --import tsx --test src/fabric/functions/validate.function.test.ts` → **66/66 pass** (both db-adapter cases green).

Test-fixture-only change; no changeset, consistent with the sibling `fix(cli): validate fixtures …` commits.

### DIFF
--- a/packages/cli/src/fabric/functions/validate.function.test.ts
+++ b/packages/cli/src/fabric/functions/validate.function.test.ts
@@ -683,7 +683,8 @@ describe('pikku fabric validate', () => {
         await makeValidProject(tmp)
         await writeFile(
           join(tmp, 'packages', 'functions', 'src', 'services.ts'),
-          "import { Kysely } from 'kysely'\nimport { PostgresDialect } from 'kysely'\n",
+          "import { Kysely, PostgresDialect } from 'kysely'\n" +
+            'export const db = new Kysely({ dialect: new PostgresDialect({}) })\n',
           'utf8'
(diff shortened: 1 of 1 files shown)


yasserf:
Superseded by 34accf48e on #1009, which fixes the same `services-wrong-db-adapter` fixtures *and* adds the negative case (a bare `import` that constructs nothing must not be flagged) — 67/67 pass. Both edit the same test file, so keeping this open would only conflict. Closing in favour of the fix riding with the catalogue PR.

PR #1020, opened 2026-07-22
fix(queue-pg-boss): eliminate head-of-line blocking (localConcurrency, not batchSize)

yasserf:
## Problem

The pg-boss worker mapped pikku's requested parallelism to pg-boss **`batchSize`**. pg-boss's batch handler receives the whole fetched batch and only fetches again once the handler resolves — and the handler `Promise.all`s the batch. So a single long-running job:

- blocks every **sibling** job in its batch until it finishes, and
- stalls the **next fetch**, so nothing queued behind it starts either.

On a shared queue that's classic **head-of-line blocking**. We hit this in production on the workflow-orchestrator queue: one long-running workflow step pinned all 10 batch slots for tens of minutes, starving short operational workflows (project deletes, deploys) queued behind it. Raising `batchSize` makes it *worse* — a bigger batch just freezes more siblings behind the slow one.

## Fix

Map parallelism to pg-boss **`localConcurrency`** with **`batchSize: 1`** instead. `localConcurrency` spawns N independent workers that each fetch and process a single job and refill the instant they finish — so a slow job never holds up the others. Same effective concurrency, no head-of-line blocking.

The existing `batchSize` worker-config knob is preserved (its documented meaning is "process concurrently in parallel"); it now drives `localConcurrency`. Default stays 10.

## Test

`mapPikkuWorkerToPgBoss` unit tests assert the mapping (`{ localConcurrency: N, batchSize: 1 }`) and `pollInterval` ms→s conversion. `tsc` clean.

### DIFF
--- a/.changeset/pgboss-worker-hol-blocking.md
+++ b/.changeset/pgboss-worker-hol-blocking.md
@@ -1,0 +1,18 @@
+---
+'@pikku/queue-pg-boss': patch
+---
+
+Stop one slow job from holding an entire queue (head-of-line blocking).
+
+The worker mapped pikku's parallelism to pg-boss `batchSize`, whose handler
+receives the whole fetched batch and only fetches again once that handler
+resolves. Because the handler `Promise.all`s the batch, a single long-running
+job blocked every sibling in its batch *and* stalled the next fetch — so on a
+shared queue one slow job could freeze the worker and starve everything behind
+it (observed with workflow-orchestrator jobs, where a long step pinned all 10
+slots for tens of minutes).
+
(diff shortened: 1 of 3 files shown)


yasserf:
Folded into #1023 — the localConcurrency mapping is the first commit there, and #1023 adds the job-group fairness that makes a shared queue viable in the first place. One PR is easier to review than two stacked ones.

Note for anyone reading the original description: the prod starvation that prompted this was actually caused by a queue-naming regression (a stale codegen bootstrap collapsed 37 per-workflow queues onto one shared queue), not by batching. Batching made it worse, and the mapping change is still right, but it wasn't the root cause.

PR #1032, opened 2026-07-23
fix(cli): fabric validate type-checks frontends; link no longer clobbers pikkufabric.config.json

yasserf:
## Why

Two failures from the same afternoon, one deploy apart.

**1. `validate` never ran tsc.** The build container type-checks each deployable frontend and aborts the deploy if it fails. `fabric validate` approximated that with ~1900 lines of structural heuristics (several literally commented "type-checks locally but fails in the clean build container") but never invoked the compiler. A `TS2353` on a Mantine `component={Link}` sailed past validate and burned a deploy eight minutes in.

**2. `link`/`init` clobbered the project config.** `writeProjectConfig` wrote whatever object it was handed; both callers hand it `{ projectId }`. That silently deleted `frontends`, so the *next* deploy built and type-checked no frontend at all — succeeding, with nothing shipped. The cause is separated from the symptom by however long it takes to deploy again.

## What

- `packages/cli/src/fabric/lib/frontend-typecheck.ts` — runs each deployable frontend's own `tsc` script (falls back to `tsc --noEmit`), via the package manager the project declares, and parses tsc's diagnostic lines out of the runner's chatter.
- `validate` reports them as errors with the raw diagnostics and a reproduce line. `--skip-typecheck` for structural-only.
- `writeProjectConfig` merges into the existing file (unknown keys preserved) instead of replacing it.

## Verified

Against a real Fabric project: clean tree → no finding; one injected type error → `ok = false` with the exact tsc diagnostic, 1.4s incremental.

## Summary by CodeRabbit

* **New Features**
  * `fabric validate` now performs frontend TypeScript type-checking for every deployable frontend before deployment.
  * Added `--skip-typecheck` to run structural checks only.
* **Bug Fixes**
  * `fabric link` and `fabric init` now merge configuration instead of overwriting it, preserving existing frontend and deployment settings.
  * Malformed or missing deployable `frontends` are reported as findings/errors during validation (rather than failing unexpectedly).
  * Invalid configuration files now warn and continue instead of stopping the command.

### DIFF
--- a/.changeset/fabric-validate-typecheck.md
+++ b/.changeset/fabric-validate-typecheck.md
@@ -1,0 +1,13 @@
+---
+'@pikku/cli': patch
+---
+
+`fabric validate` now type-checks every deployable frontend, the same compile the
+build container runs before it will deploy, so a type error fails locally in
+seconds instead of minutes into a deploy. Pass `--skip-typecheck` for the
+structural checks alone.
+
+Also fixes `fabric link` / `fabric init` clobbering `pikkufabric.config.json`:
+they wrote a fresh `{ projectId }` object, silently deleting `frontends`,
+`production` and `apiUrl`. Losing `frontends` means the build container deploys
+no frontend at all, long after the link that caused it. The config is now merged.
--- a/.changeset/fabric-validate-typecheck.md
+++ b/.changeset/fabric-validate-typecheck.md
@@ -7,7 +7,13 @@ build container runs before it will deploy, so a type error fails locally in
 seconds instead of minutes into a deploy. Pass `--skip-typecheck` for the
 structural checks alone.
 
+Declared frontends are shape-checked before they are used, so a malformed
+`frontends` entry is reported as a finding instead of throwing part-way through
+validation, and a deployable frontend whose directory does not exist is an error
+even in a project with no `apps/` directory.
+
 Also fixes `fabric link` / `fabric init` clobbering `pikkufabric.config.json`:
 they wrote a fresh `{ projectId }` object, silently deleting `frontends`,
 `production` and `apiUrl`. Losing `frontends` means the build container deploys
-no frontend at all, long after the link that caused it. The config is now merged.
+no frontend at all, long after the link that caused it. The config is now merged,
+and a config that cannot be read at all is left alone rather than overwritten.
--- a/packages/cli/src/fabric/fabric-commands.ts
+++ b/packages/cli/src/fabric/fabric-commands.ts
@@ -73,6 +73,13 @@ export const fabricCommands = defineCLICommands({
     render: renderValidate,
     description:
       'Check the project structure for fabric compatibility — prints all missing or misconfigured items with fix hints',
+    options: {
+      skipTypecheck: {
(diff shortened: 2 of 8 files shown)


yasserf:
Addressed in 49f85d1c9.

**Fixed**

- **Unvalidated frontend entries (critical)** — entries are shape-checked once, up front, before anything dereferences them. A `null` entry → `frontend-entry-invalid-<slug>`, a non-string/empty `cwd` → `frontend-cwd-invalid-<slug>`, a non-object `frontends` → `frontends-invalid`. The same latent throw existed in the earlier `apps/` block (`fe.cwd?.replace()`), which now consumes the narrowed list instead of re-reading raw JSON.
- **Missing directory silently dropped (major)** — the `existsSync` filter is gone. A deployable frontend whose directory does not exist is now `frontend-cwd-missing-<slug>` regardless of whether `apps/` exists; the check moved out of the `apps/`-gated block for exactly that reason.
- **Reproduce line (minor)** — `typeCheckFrontends` returns the invocation it chose (`command` on the result), and the finding renders that. So a yarn project gets `yarn run tsc` and a fallback gets `npx tsc --noEmit`, instead of `npm run tsc` in every case.
- **Silent non-zero exit (critical)** — `parseTscErrors('')` returned `[]`, which reads as "compiled fine". It now returns a fallback failure line, so a `tsc` that fails without output cannot pass validation.
- **Read failure vs parse failure (major)** — `readFile` and `JSON.parse` are separated. A permissions/IO error now throws (`refusing to overwrite it`) rather than resetting `existing` to `{}` and clobbering the file this function exists to protect. Valid-but-non-object JSON warns before it is replaced, rather than being dropped silently.
- **Verifier coverage** — 23 new tests, all failing before the change:
  - `src/fabric/lib/config.test.ts` — merge preserves `frontends`/`production`/unknown keys, partial `apiUrl` update, fresh write, unparseable rewrite, non-object rewrite, and the unreadable-config refusal (asserts the original bytes survive).
  - `src/fabric/lib/frontend-typecheck.test.ts` — diagnostic parsing vs runner chatter, tail fallback, empty-output fallback, script-vs-`--noEmit` selection across yarn/pnpm/npm, `tsconfig` skip, runner-not-on-PATH skip, and end-to-end pass/fail/silent-fail via a stub package manager.
  - `src/fabric/functions/validate.function.test.ts` — the malformed-config findings above, missing-dir without `apps/`, `--skip-typecheck` opt-out, and `deploy: false` exclusion.

**Not applied**

- **"Use Yarn for the monorepo command path"** — this code does not run in this monorepo. `fabric validate` runs inside a *user's* Fabric project and has to invoke whatever package manager that project declares; hardcoding yarn would run the wrong manager (or none) in a bun/pnpm/npm project, and would resolve a different `typescript` than the build container does. `AGENTS.md:59` scopes that rule to "all commands in **this monorepo**". The immediately preceding commit on `main` (`9a9ed6f16 fix(console): detect the project package manager instead of defaulting to yarn`) is the same fix in the other direction. The related complaint — that the *reproduce hint* claimed npm — is fixed above.

PR #1038, opened 2026-07-26
feat(cli): schema types from migrations, and a drift check

yasserf:
Two commands, both built on the scratch-database mechanism the CLI already has (`withScratchPostgresDatabase` — in-memory PGlite for Postgres, `:memory:` for SQLite, needing no `CREATEDB` privilege anywhere).

## `pikku db codegen`

Regenerates the database types from the migration files alone, without connecting to a database.

`db migrate` can only emit types *after* it has migrated the real database, which forces codegen to run late. On a deploy that is the wrong order: the build step that needs the table zod (a function schema built from `#pikku/db/zod.gen.js`) runs long before migrations do, so it sees a stale or missing schema and fails on the VM while passing locally. The workaround has been to hand-mirror columns next to a table that already declares them.

`db codegen` applies the same migrations to a throwaway database and introspects that, so `pikku all` can be handed an accurate schema on a machine with no database configured or reachable at all.

The generated types describe what the migrations *define*, which is the contract. Introspecting a live database additionally picks up whatever has drifted into it; that difference is the point, and it is what the second command reports.

## `pikku db check`

Reports how the configured database differs from the schema its migrations define — the question nobody can otherwise answer without going and looking by hand.

The two halves are deliberately asymmetric:

- **Missing** tables/columns mean the database is behind. The fix only ever adds, so this **fails** the command and points at `db migrate`.
- **Extra** tables are reported and never acted on. Something created them outside the migration history, and no migration can know whether they hold data worth keeping.

Comparison is on the fully-qualified name, so a second copy of a table in the wrong schema (`public.orders` shadowing `app.orders` — what a runtime that forgot to qualify its DDL leaves behind) is reported rather than mistaken for the table the migrations created.

## Verification

Run against a real project with 76 migrations:

- `db codegen` with **no database configured**: applied all 76 to a scratch Postgres and regenerated `schema.gen.ts` + `zod.gen.ts` (84 tables). Output is byte-identical to the live-database run for every migration-defined table.
- `db check` against that project's live Postgres correctly identified 4 tables a runtime bootstraps at boot and 5 unqualified duplicates sitting in `public` alongside their `app.` originals — a real bug the project did not know it had.

442 tests pass, including four new ones covering scratch codegen on both dialects, a database behind its migrations, an unrecorded table, and the schema-shadow case.

## Summary by CodeRabbit

* **New Features**
  * Added `pikku db codegen` to generate database types and schemas from migrations without connecting to the configured database.
  * Added `pikku db check` to compare the configured database with migration-defined schemas.
  * Reports missing tables or columns and recommends running `pikku db migrate`.
  * Reports extra database tables without failing or removing them.
* **Documentation**
  * Clarified database schema comparison, migration behavior, and release metadata.

### DIFF
--- a/.changeset/mighty-cases-shake.md
+++ b/.changeset/mighty-cases-shake.md
@@ -1,0 +1,11 @@
+---
+'@pikku/cli': minor
+---
+
+Add `pikku db check` — report how the configured database differs from the schema its migrations define.
+
+Answers the question nobody can otherwise answer without going and looking by hand: does this database still match what we wrote down? It applies the migrations to a throwaway database (the same scratch mechanism `db codegen` uses) and diffs that against the real one.
+
+The two halves are deliberately asymmetric. Missing tables and columns mean the database is behind — the fix only ever adds, so the command fails and tells you to run `db migrate`. Tables the migrations never mention are reported but never fail and never dropped: something created them outside the migration history, and no migration can know whether they hold data worth keeping.
+
+Comparison is on the fully-qualified name, so a second copy of a table in the wrong schema (`public.orders` shadowing `app.orders` — what a runtime that forgot to qualify its DDL leaves behind) is reported rather than mistaken for the table the migrations created.
--- a/.changeset/quiet-rules-sniff.md
+++ b/.changeset/quiet-rules-sniff.md
@@ -1,0 +1,9 @@
+---
+'@pikku/cli': minor
+---
+
+Add `pikku db codegen` — regenerate the database types from the migration files without connecting to a database.
+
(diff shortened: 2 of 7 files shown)


yasserf:
Superseded by #1042, which is merged.

That PR carries both of this branch's commits verbatim — `db codegen` and `db check` — and builds on them: `db baseline`, `db export`, and the `@pikku/kysely` declared-schema refactor that lets `db check` attribute boot-created tables to the runtime that declares them.

Merging `origin/main` into this branch and resolving every conflict produces a tree byte-identical to main, so there is nothing left here to land.

Issue #1039, opened 2026-07-26
perf(workflow): optimise the workflow service — indexes, replay snapshot, step-transition round-trips


yasserf:
The workflow service has accumulated a set of hot-path inefficiencies. Audit findings, ranked:

### Tier 1
1. **No indexes on any workflow table.** `KyselyWorkflowService.init()` creates 4 tables with zero indexes; every other Kysely service in the repo creates them. `workflow_step_history.workflow_step_id` is unindexed and is the hottest table: correlated `count(*)` on every `getStepState`, `ORDER BY created_at DESC LIMIT 1` on every running/result/error write, and 4 correlated subqueries per row in `getRunSteps`.
2. **Replay is O(N^2) in DB round-trips.** Every orchestrator tick replays the body from the top; each step does its own `getStepState`. N steps x N ticks. Fix: one snapshot load per replay.
3. **Every step transition is 3 non-transactional round-trips.** UPDATE step -> SELECT latest history -> UPDATE history. A crash between writes leaves step `succeeded` while history says `running`, corrupting `getRunHistory` and the timeline.

### Tier 2
4. `updateRunState` is a lossy read-modify-write over the whole `state` blob — concurrent `graph.setState` writes lose each other.
5. `getRun` re-fetched 3-5x per step; `resumeWorkflow` already takes an optional `workflowName` that call sites don't pass.
6. `getAIGeneratedWorkflows()` used as a point lookup — fetches and parses every dynamic workflow to `.find()` one.
7. `JSON.parse(JSON.stringify(payload))` on every dispatch.
8. Fixed 1000ms / 500ms polling in `runToCompletion` and the inline child-workflow wait.

### Tier 3
9. Per-run state leaks: `stepLineage` never cleared on run end; `runActors` only cleared on the inline branch.
10. Scenario surface (~800 lines) lives inside the durable-execution engine; `createWorkflowWire` builds all 12 closures per run even for plain workflows.
11. The `xxx`/`xxxImpl`/`safeMirror` triple is copy-pasted 12 times.

yasserf:
Shipped in #1040 (merged 2026-07-30). Verified against current `main`:

1. Indexes — `packages/services/kysely/src/schema/workflow.schema.ts:121,127,133,139,145` (`idx_workflow_runs_status_created`, `idx_workflow_runs_workflow_created`, `idx_workflow_step_run_status`, `idx_workflow_step_history_step`, `idx_workflow_versions_source_status`).
[...]

PR #1040, opened 2026-07-26
perf(workflow): index the tables, cut the reads, and fix two lost-state bugs

yasserf:
Works through all eleven items of #1039.

## Tier 1 — the ones that cost the most

**No indexes on any workflow table.** Every step read, history walk and orchestrator tick was a sequential scan. Five indexes now cover the columns the engine actually queries by, created through a `createIndexSafe` that tolerates an existing index.

**Replay was O(N²) in round trips.** A replay walks the DSL body from the top and each step it passes asked for its own row — N reads per replay, N² over a run. `listStepStates(runId)` reads them all once at the start of a replay pass and serves the walk from that snapshot. Backends without a bulk read return `null` and fall back to the old path. The two lock-protected sites (`suspendStep`, `approvalStep`) and the worker entry points deliberately do *not* use the snapshot.

**A step transition was three non-transactional round trips.** A crash between them left a step row saying `succeeded` whose history row still said `running`, which silently corrupts `getRunHistory` and every timeline built on it. Both halves are now one transaction. The history row is found by attempt number rather than by sorting on `created_at` — two attempts can land in the same millisecond — which needed a `current_attempt` column, added with an in-memory backfill for existing databases (MySQL rejects the correlated subquery that would have done it in SQL).

## Tier 2

- `updateRunState` was a lossy read-modify-write; it now goes through a dialect `jsonSetState` hook.
- `getRun` was re-read 3–5x per step; one read per orchestrator tick now, cached for the replay.
- `getDynamicWorkflow(name)` replaces reading and parsing every AI-generated workflow in the deployment to `.find()` one by name. Implemented per backend.
- Dispatch no longer JSON round-trips every step payload. SQS, BullMQ and pg-boss all serialise on the way to the wire, so it was waste on every production path; its one real effect was giving the in-process dev queue the isolation a real backend provides, so that moved into `InMemoryQueueService`.
- Waiting on a run backs off instead of polling flat. `pollIntervalMs` is now a ceiling, not a cadence: it starts at 10ms and widens by 1.6x. An 8-read run in the new suite drops from 7.0s to 0.43s.

## Tier 3

- Per-run state (`stepLineage`, `runActors`, inline flags, ordinals) collapsed into one `RunContext` map, released when a run reaches a terminal state — `suspended` deliberately excluded, since a suspended run can still be resumed.
- The scenario assertion surface (`expectEventually`, `expectError`, `expectService`, `runScheduledTask`) moved out of the engine into `scenario-wire.ts`.
- The thirteen copy-pasted `x` / `xImpl` / `safeMirror` triples became one `mirrored(write, mirror)`.

## Two defects found while porting to the other backends

**Redis lost run state.** State was one JSON blob read-modified-written on every `setState`, so two parallel branches setting different variables overwrote each other — twenty concurrent writers left one survivor. It is a field per variable now. `getRunState` still reads the old blob underneath, so a run in flight across the deploy keeps what it had.

**Mongo never wrote scheduled history.** `setStepScheduled` wrote the step row and nothing else, so a queued step's history stayed at `pending` and read as never dispatched. Its other three transitions also found the newest history document and then updated it by `_id`; `findOneAndUpdate` with the same sort does it in one.

## Verification

1618 tests green: core 1349, kysely 177, mongodb 76, redis 16. All five touched packages typecheck clean.

New suites, each proven red before the change: workflow table indexes, step-transition round trips and consistency, attempt counting, run-state writes, upgrading a pre-attempt database, dynamic workflow lookup, replay snapshot, run context lifetime, run polling backoff, dispatch payload, redis run state, mongo step history.

Two commits are behaviour-preserving refactors (`mirrored`, the scenario extraction) so their tests are characterisation rather than red-first. Both were mutation-checked: the mirror suite catches dropping a mirror call, mirroring before the write, and letting a mirror failure escape. `expectError` and `expectService` had no coverage at all before this and now do.

## Not done

Mongo's step and history writes are still two operations — multi-document transactions need a replica set, and that would break standalone deployments and the package's own `mongodb-memory-server` tests. Redis's step transitions likewise still read `attemptCount` before writing it.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic dynamic workflow selection, choosing the newest active version.
  * Improved workflow state handling across retries, replays, and concurrent updates.
  * Added safer per-variable run-state storage while preserving compatibility with existing runs.

* **Bug Fixes**
  * Fixed missing or inconsistent step history during transitions and retries.
  * Reduced replay reads and improved polling responsiveness with bounded backoff.
  * Improved JSON state updates across supported databases and preserved queued payload data accurately.

* **Chores**
  * Added database migration support and indexes for improved workflow performance.
  * Expanded automated testing in continuous integration.

### DIFF
--- a/.changeset/olive-pugs-shake.md
+++ b/.changeset/olive-pugs-shake.md
@@ -1,6 +1,8 @@
 ---
 '@pikku/core': patch
 '@pikku/kysely': patch
+'@pikku/kysely-mysql': patch
+'@pikku/kysely-postgres': patch
 '@pikku/mongodb': patch
 '@pikku/redis': patch
 '@pikku/cloudflare': patch
--- a/.changeset/olive-pugs-shake.md
+++ b/.changeset/olive-pugs-shake.md
@@ -35,3 +35,17 @@ dispatched.
 Also: dispatch no longer JSON round-trips every step payload before handing it
 to a queue that serialises it anyway — the in-process dev queue, which is the
 only one that was relying on it, does it itself now.
+
+Two more defects. A transition whose step had no live attempt wrote its status
+to the step row and silently nothing to history — the exact divergence the
+transaction exists to prevent — and now repairs the step and writes the
+missing row. And resolving a dynamic workflow was non-deterministic on all
+three backends: a name can hold several active versions, and none of them
+ordered the candidates, so which one ran could change between two calls
+reading identical data. The newest version wins, with the graph hash breaking
+a tie.
+
+The two attempt columns and the five indexes are declared in the workflow
+schema, so a fresh database gets them at boot. An existing one gets them from
+a migration — `pikku db generate` writes the declaration down — rather than
+from DDL issued at boot.
--- a/.changeset/olive-pugs-shake.md
+++ b/.changeset/olive-pugs-shake.md
@@ -1,0 +1,37 @@
+---
+'@pikku/core': patch
+'@pikku/kysely': patch
+'@pikku/mongodb': patch
+'@pikku/redis': patch
+'@pikku/cloudflare': patch
+---
+
+Make the workflow service cheaper to run, and fix two ways it lost state.
+
+The SQL workflow tables had no indexes at all, so every step read, every
+history walk and every orchestrator tick was a sequential scan; five indexes
+now cover the columns the engine actually queries by. A replay used to ask for
+each step's row individually — O(N) reads per replay, O(N^2) over a run — and
+now takes one read of the run's steps and serves the walk from it. A step
+transition wrote the step row and its history row as two separate statements,
+so a crash between them left a step saying `succeeded` whose history still said
+`running`; both halves are now one transaction, and the history row is found by
+attempt number rather than by sorting on `created_at`, which two attempts can
+share. Resolving a dynamic workflow read and parsed every AI-generated workflow
+in the deployment to `.find()` one by name; it is a point lookup now.
+
+Waiting on a run no longer polls at a fixed interval. `pollIntervalMs` became a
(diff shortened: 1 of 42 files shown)


yasserf:
Followed up on the Postgres and SQLite paths specifically.

**Postgres had no tests at all.** Its dialect SQL diverges from the base exactly where it is easiest to get wrong — `jsonb ||` in place of `json_set`, its own casts, advisory locks — and none of it had ever executed in CI, so the `jsonSetState` override this PR added was shipped unverified.

It runs against real Postgres now. `@electric-sql/pglite` is Postgres compiled to WASM, so `jsonb`, transactional DDL, advisory locks and the real SQLSTATE codes behave as they do on a server, and it needs no Docker. A small Kysely driver over it lives in the test file, so it stays out of `dist`. 17 tests cover schema and index creation, the jsonb state merge, step transitions and retries, both advisory locks, and an upgrade from a database written before the `attempt` columns existed.

Since the suite is a refactor’s safety net, it was mutation-checked: dropping the Postgres `jsonSetState` override, casting its value to `text` instead of `jsonb`, and leaving the path segment quoted each turn six tests red.

**SQLite** needed no harness — `SQLiteKyselyWorkflowService` is an empty subclass, so the existing `better-sqlite3` suite already exercised it. It did not, however, assert what `json_set` does to values, so it gains the same matrix Postgres now has: an empty array, zero, `false` and `null` each have to survive the round trip as themselves, and a key containing a dot or a quote has to stay one literal key rather than being read as a nested path. That last test fails if the JSON path stops being quoted, which nothing caught before.

180 green in `@pikku/kysely`, 17 in `@pikku/kysely-postgres`, 1349 in `@pikku/core`; all four dialect packages typecheck.

PR #1042, opened 2026-07-27
feat(db): declared schemas — one source of truth per schema, and one place that writes it down

yasserf:
Supersedes #1038, whose two commits are the first two here.

The runtime's tables stop being a side effect of which services happened to start, and `pikku db generate` stops knowing about only one schema.

## `@pikku/kysely` — one declaration instead of `init()` DDL

The tables the runtime needs (workflow, AI, scopes, secrets, credentials, webhooks, channels, deployments, sessions) were created by `createTable(...).ifNotExists()` inside each service's `init()`. A table existed if and only if some service was constructed, in whatever order, with whatever schema qualifier that call site passed — in one deployment that produced a duplicate set of AI tables in `public` shadowing the intended ones in `app`.

`pikkuSchemas` is now the single declaration. `applyPikkuSchemas` materializes it, `compilePikkuSchemas` renders it as SQL, and every service's `init()` goes through `ensurePikkuSchema`, which looks before it creates: all present is a no-op, none present creates, *some* present throws rather than filling in the rest. `.ifNotExists()` had turned every failure into a silent success. ~615 lines of duplicated DDL went with it.

Auth is a declared prerequisite rather than an assumption: `pikkuUserRole`/`pikkuUserScope` reference `user.id`, and `userId` now takes its type from the column it references. The hand-written DDL said `text`, which postgres rejects against the `uuid` or identity `integer` Better Auth emits under `generateId` — so those tables had never actually been created on such a project.

## `@pikku/cli` — the schema-source registry

- **`pikku db codegen`** — regenerate database types from the migration files, no database connection.
- **`pikku db check`** — diff a database against its migrations, and tell a table the runtime created apart from one nobody can explain.
- **`pikku db generate`** now walks a registry — Better Auth, the runtime declaration, then each wired addon — and writes one migration per source in dependency order. Fully covered is a no-op; nothing covered emits the source's own SQL verbatim (it carries the indexes and constraints a column diff cannot see); partially covered emits `ALTER TABLE … ADD COLUMN` deltas.
- **`pikku db baseline`** — record pending migrations as applied without running them, for the database that already contains what they describe. A separate command rather than a flag on `migrate`, because a flag pasted into a deploy script would silently stop applying migrations forever. It refuses unless the database really is up to date.
- **`pikku db export`** — an addon publishes its schema to `.pikku/db/pikku-db-meta.gen.json`; `db generate` folds each *wired* addon's schema into the consumer's history. A `wireRemoteAddon` addon contributes nothing.

## `verifiers/db-schema`

The unit tests hand-wrote the addon artifact, so no addon ever ran `pikku db export`, no `wireAddon` declaration reached `db generate` through the inspector, and no generated migration was ever applied — and the auth source ran against a bare config, a four-table core a generator could hardcode and still pass.

The verifier runs the real CLI binary against a real database, on both dialects. Better Auth is configured with `organization()`, `admin()` and `twoFactor()`, so the auth source has to be what Better Auth materializes — and because `admin()` adds *columns* rather than tables, a generator that handles table creation but not column addition fails. Beside it is a real addon that ships a table and is discovered because it is wired. Thirteen steps: generate → migrate → check → generate again → runtime → baseline → baseline refused.

It found two bugs on its own runs:

- `db generate` crashed with `ENOENT` on a project with no `db/<dialect>` directory — the first run on a new project, the exact case the command exists for.
- **`sql_migrations` leaked into postgres introspection.** SQLite had always hidden the migrator's own tracking table; postgres reported it like any other. So an addon exported from a migrated postgres database published `sql_migrations` as one of *its* tables, the consumer has it too, and `db generate` read the addon as partially covered — writing column deltas instead of the addon's SQL, dropping its primary key and unique index. The table was created, and created wrong.

CI runs the sqlite pass in the `verifiers` matrix and the postgres pass in a job with a postgres service of its own (the run drops and recreates `public`).

## Release

Every changeset here is a **patch**. `@pikku/cloudflare` peer-depends on `@pikku/kysely`, and changesets majors a peer dependent on any non-patch bump — which is how it was published as 1.0.0 once already (b3880e531, unpublished and rolled back to 0.12.13). `changeset status` reports patches only: `@pikku/cli` and `@pikku/kysely`, no minor, no major, `@pikku/cloudflare` not bumped.

## What the verifier caught on a clean checkout

A `wireRemoteAddon` addon must contribute nothing. The unit test passes `remote: true` in by hand, so `verifiers/db-schema` now ships a second addon — `notes`, published through the same `pikku db export` channel, as complete an artifact as the one that *is* folded in — and wires it remotely. Wired as `wireAddon` instead, six assertions fail and the table is created; wired remotely, no fourth migration is written and `notes` is absent from the database after `db migrate`.

Clearing `src/scaffold` alongside `.pikku` so the run starts where a checkout starts then failed immediately, and explained why CI failed where local did not. The auth factory's services reach the generated services map by being stamped onto the auth handler's function meta — and on the first run in a clean checkout that handler's file does not exist yet. Nothing to stamp, nothing to aggregate: `kysely` came out unused, `RequiredSingletonServices` typed it optional, and `tsc` rejected the runtime check. The second run passed. CI is always the first run, and the same map decides what the deployed auth worker keeps.

`aggregateRequiredServices` now reads those services from the auth definition, which is inspected from hand-written source on every pass.

## Summary by CodeRabbit

* **New Features**
  * Added `pikku db check`, `codegen`, `baseline`, and `export` commands for schema validation, type generation, migration tracking, and addon schema publishing.
  * Database migrations now support runtime, authentication, and addon schemas.
  * Added clearer schema prerequisite and drift reporting.
* **Bug Fixes**
  * Improved consistent authentication service inclusion across clean and incremental builds.
  * Prevented partial schema creation and corrected runtime table definitions and column types.
* **Tests**
  * Added comprehensive SQLite, PostgreSQL, addon, migration, and schema verification coverage.


yasserf:
@coderabbitai review

yasserf:
Worked through the review — four applied in `1018b43b8`, two declined.

**Applied**

- **`applyPikkuSchemas` atomicity.** Right, and it contradicted the function's own doc ("or a half-applied database to clean up") — the prerequisite check only covers the failures it can foresee. The statements now run in one transaction. There's a test for it (`a statement failing part way takes the ones before it with it`), and I checked it fails without the transaction rather than passing for free.
- **`KyselyWorkflowMirror.init` doc.** Stale in the exact way this branch exists to fix: it still promised `ifNotExists`, while `ensurePikkuSchema` deliberately throws on a half-applied schema. Rewritten to say what it does.
- **SQLite `TEXT PRIMARY KEY` nullability.** Applied to both verifier addons (`labels` as well as `notes`) — same latent hole, and `labels` is the one that actually migrates into the consumer's database.
- **`auth.wiring.ts` reading `process.env` directly.** Now `config.postgresUrl`. `createConfig` already reads that env var; reading it twice is two sources of truth for the dialect.

**Declined**

- **Workflow hardening (`permissions:`, `persist-credentials: false`, pinned `postgres` digest).** No job in `develop.yml` sets `permissions:` or `persist-credentials`, and the two pre-existing `image: postgres` services are equally unpinned. Hardening one of eighteen jobs because it happens to be new leaves the file inconsistent and the repo no safer — this wants a repo-wide pass, in its own PR.
- **`db generate` not diffing indexes or functions.** That's the documented design, not an oversight. A source nothing covers is emitted as its own SQL verbatim *precisely* so indexes and constraints survive; only the partially-covered case falls back to column deltas, and that case emits a `REVIEW:` note saying a column list carries no indexes or foreign keys. Index-level diffing is a feature on top, not a fix to this one.

Both dialects of the db-schema verifier re-run green, and `@pikku/kysely` is 168/168.

yasserf:
Amending the "declined" above: `8a2b213f5` does take one piece of the workflow finding after all, because the flake it prevents kept hitting this PR.

The failures were `registry-1.docker.io … Client.Timeout exceeded while awaiting headers` during GitHub's `Initialize containers` step — a connection to Docker Hub that never opened, three retries deep, before any step in the job runs. Not a 429, so not the pull-limit story, and not something a retry inside the job can reach.

`postgres` and `redis` now come from `mirror.gcr.io/library/…`, Google's read-through cache of Docker Hub's official images: no auth, no sync job, nothing to own. Both are pinned on the way past, which is the part of the original finding this shares. `softwaremill/elasticmq-native` isn't an official image so the mirror doesn't carry it — it stays on Docker Hub until it gives us a reason not to.

Still declined: `permissions:` and `persist-credentials: false`. No job in either workflow sets them, and adding them to one is a repo-wide policy question wearing a one-job disguise.

PR #1045, opened 2026-07-27
fix(kysely): ensurePikkuSchema on a schema-bound connection

yasserf:
## The bug

`ensurePikkuSchema` read the table name out of its own compiled DDL with a regex that matched the **first** quoted identifier:

```ts
/^\s*create table "([^"]+)"/i
```

On a connection bound to a schema that DDL is `create table "app"."workflow_runs"`, so it captured `app` — never a real table name. Every table therefore read as missing, and it issued a bare `create table` that the database rejected:

```
PostgresError: relation "workflow_runs" already exists
```

On every boot, forever. This crash-looped a production backend 82 times after upgrading to `0.13.4`, which is where the introspect-then-create replaced the old `.ifNotExists()` DDL.

## The fix

Parse the optional schema half, and match on the pair when the DDL is qualified:

```ts
const CREATE_TABLE = /^\s*create table (?:if not exists\s+)?(?:"([^"]+)"\.)?"([^"]+)"/i
```

`declaredTables` now returns `DeclaredTable[]` (`{ schema?, name }`) instead of `string[]`, and the lookup indexes introspection both ways. A qualified declaration matches on `schema.name`; an unqualified one falls back to the bare name, since such a statement resolves against a `search_path` that isn't knowable from here. `isPresent` branches on the **declaration**, not on what introspection happens to report — which matters, because `SqliteIntrospector` returns `schema: undefined`. Half-applied and boot-race paths share the predicate, and error messages name the schema.

## Second commit: the sqlite case

While verifying sqlite still worked, `withSchema` turned out to be broken there for an unrelated reason. Not create-table — raw sqlite accepts `create table "main"."foo"` fine. It's the foreign key:

```
qualified create:       OK
qualified references:   FAILED: near ".": syntax error
unqualified references: OK
```

`withSchema` qualifies reference targets too. Postgres **needs** that (an unqualified reference resolves against `search_path`, which a schema-bound connection doesn't control — the mirror image of the bug above); sqlite **forbids** it, because a foreign key there can only point inside the same database, so there is no qualifier to give. One declaration can't spell both.

Rather than rewrite compiled SQL per engine — more string-munging of exactly the kind that caused the first bug — the failure is made legible. `applyPikkuSchemas` catches per statement and, when the SQL has a qualified `references` **and** the engine said "syntax error", reports the schema, both ways out, and keeps the engine error as `cause`. The double gate means a qualified reference that fails for any other reason still surfaces as itself. No dialect sniffing, so it holds for any sqlite driver (bun-sqlite, libsql) rather than only the one whose adapter class we'd have recognized.

## Tests

New `ensurePikkuSchema — a connection bound to a schema` suite. The discriminating cases need a stubbed Postgres introspector, since sqlite has one schema and can't express them:

- `declaredTables` reports the schema for qualified DDL, `undefined` otherwise
- tables present under `app` on a `withSchema('app')` connection → `present` (before: `created`, then the real database throws "already exists")
- the same names under `public` → still `created`; a different table answers nothing
- an unbound connection still matches on the bare name
- the half-applied error names the schema
- the qualified-foreign-key message, driven against a real in-memory sqlite, asserting the surviving `cause`

`174 pass, 0 fail`; `tsc --noEmit` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed schema detection for database connections using schema-qualified tables.
  * Prevented repeated attempts to create tables that already exist.
  * Improved error messages for partially applied schemas with qualified table names.
  * Enhanced SQLite errors for unsupported schema-qualified foreign-key syntax, including recommended guidance and the original database error.
  * Improved handling of tables with identical names across different schemas.

### DIFF
--- a/.changeset/kysely-ensure-schema-qualified.md
+++ b/.changeset/kysely-ensure-schema-qualified.md
@@ -15,3 +15,10 @@ The schema half of the name is now parsed, and the lookup matches on the pair
 when the DDL is qualified (falling back to the bare name when it is not, since
 an unqualified statement resolves against a `search_path` that is not knowable
 from here). Error messages name the schema too.
+
+Also: applying a schema over a `withSchema(...)`-bound **sqlite** connection now
+explains itself. `withSchema` qualifies foreign key targets along with everything
+else — which postgres requires and sqlite refuses, since a `REFERENCES` clause
+there takes a bare table name — and all the engine says about it is
+`near ".": syntax error`. The failure is now reported with the schema that
+produced it and what to do instead, with the engine error kept as the `cause`.
--- a/.changeset/kysely-ensure-schema-qualified.md
+++ b/.changeset/kysely-ensure-schema-qualified.md
@@ -1,0 +1,17 @@
+---
+'@pikku/kysely': patch
+---
+
+fix: `ensurePikkuSchema` now sees tables on a `withSchema(...)`-bound connection
+
+It read the table name out of its own compiled DDL with a regex that matched the
+first quoted identifier. On a connection bound to a schema that DDL is
+`create table "app"."workflow_runs"`, so it captured `app` — never a real table
+name — concluded every table was missing, and issued a bare `create table` that
+the database rejected with `relation "workflow_runs" already exists`, on every
+boot.
+
+The schema half of the name is now parsed, and the lookup matches on the pair
+when the DDL is qualified (falling back to the bare name when it is not, since
+an unqualified statement resolves against a `search_path` that is not knowable
+from here). Error messages name the schema too.
--- a/packages/services/kysely/src/schema/index.ts
+++ b/packages/services/kysely/src/schema/index.ts
@@ -143,10 +143,48 @@ export const applyPikkuSchemas = async (
     const bound = bind(trx)
     for (const schema of schemas) {
       for (const statement of schema.statements) {
-        await statement(bound, types).execute()
+        const query = statement(bound, types)
+        try {
+          await query.execute()
+        } catch (error) {
+          throw explain(error, query.compile().sql, schema)
+        }
       }
     }
   })
+}
+
+/** `references "app"."workflow_runs"` — a foreign key onto a qualified table. */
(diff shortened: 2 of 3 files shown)


yasserf:
Both findings were valid — fixed in 580423b30. Thanks, these were good catches.

**1. `explain()` gate.** Agreed on the risk, though I went a different route than suggested. Gating on a driver's `code === 'SQLITE_ERROR'` trades one wrong answer for another: better-sqlite3, `node:sqlite`, `bun:sqlite` and libsql all disagree about how to label errors, so requiring a specific `code` would make the message *disappear* on the drivers that don't set it — reintroducing the opaque `near ".": syntax error` this exists to replace.

What they do agree on is passing the engine's message through verbatim. So the gate is now sqlite's own phrasing:

```ts
const SQLITE_REJECTED_QUALIFIER = /near "\.":\s*syntax error/i
```

Postgres phrases the same class the other way round — `syntax error at or near "."` — so this cannot fire on it, and it holds for any sqlite binding. Worth noting the reachable postgres path you implied is real: `resolveRequirements` interpolates an introspected column type into the DDL verbatim, so a postgres syntax error on FK-bearing DDL is not hypothetical.

**2. Schemaless introspection.** Correct, and it's the same false negative as the original bug arrived at from the other side. `tablesOf` now carries a `schemaAware` flag (`tables.some(t => t.schema !== undefined)`) and no longer indexes `undefined.x` entries; `isPresent` requires the qualified pair only when the engine actually reports schemas, falling back to the bare name otherwise.

One correction to the failure mode as described: for schemas that *have* foreign keys, sqlite + `withSchema` never reached the "reports it missing" state — it died earlier on the qualified `REFERENCES`. The bug was reachable for the FK-free schemas (`session`, `secret`, `channel`), where sqlite accepts the qualified `CREATE TABLE` and then fails with `table already exists`. Same fix either way.

Regression tests added for both: schemaless introspection with a qualified declaration resolves to `present`, and a postgres-style syntax error on FK DDL passes through untouched (via a throwing driver). 176 pass, 0 fail; `tsc --noEmit` clean.

PR #1058, opened 2026-07-29
fix(kysely-sqlite): encode Date and JSON bind args in the libsql web dialect

yasserf:
## The bug

`LibsqlWebDialect.encodeArg` handles string / number / bigint / boolean / ArrayBuffer / Uint8Array and throws `libsql: unsupported argument type object` on anything else — including a `Date`.

The CLI's `node:sqlite` dev runtime (`db/sqlite/sqlite-kysely.ts`) does not have that gap; its `coerce()` maps `Date -> ISOString`, `boolean -> 0/1` and `object -> JSON.stringify` before binding.

So the two runtimes disagree about what a bind parameter may be, and the disagreement only shows up after deploy: a query that stamps a `createdAt` or writes a JSON column passes under `pikku dev` and throws in a Worker, where this dialect is the one in use. Reads were unaffected, so the symptom is an app where every insert and update fails while every list and get works.

Found on a deployed Fabric app: `createSession`, `createEntry` and a reminder cron were failing 100% with this error, against a stage whose read RPCs were fine.

## The fix

Two branches in `encodeArg`, mirroring `coerce()`: `Date` -> ISO-8601 text, any other plain object or array -> JSON text. Values still decode as stored (ISO string / JSON string), matching what the dev runtime writes and reads. A type the dialect genuinely cannot represent still throws.

Adds a test file (and the standard `run-tests.sh` / `test` script the package was missing) covering all four cases through Kysely with a stubbed `fetch`.

## Summary by CodeRabbit

* **New Features**
  * Date parameters are now transmitted as ISO-8601 strings.
  * Plain objects (including null-prototype records) and arrays are now transmitted as JSON strings.
  * Non-JSON-encodable values (for example `Map`, `RegExp`, and class instances) are rejected with clearer “unsupported argument type” errors.
  * Primitive value handling remains unchanged.
* **Tests**
  * Expanded and reorganized parameter encoding tests, including additional negative cases.
  * Added test commands and a shared test runner supporting standard, watch, and coverage modes.

### DIFF
--- a/.changeset/libsql-web-dialect-encodes-dates-and-json.md
+++ b/.changeset/libsql-web-dialect-encodes-dates-and-json.md
@@ -1,9 +1,12 @@
 ---
 '@pikku/kysely-sqlite': patch
+'@pikku/cli': patch
 ---
 
-`LibsqlWebDialect` now encodes a `Date` bind parameter as an ISO-8601 string and any other plain object or array as JSON, instead of throwing `libsql: unsupported argument type object`.
+`LibsqlWebDialect` now encodes a `Date` bind parameter as an ISO-8601 string and an array or plain record as JSON, instead of throwing `libsql: unsupported argument type object`.
 
-This closes a dev/prod split that broke every write in a deployed libsql app. The CLI's `node:sqlite` dev runtime already coerces dates, booleans and objects before binding them, so a query that stamps a `createdAt` or writes a JSON column passes under `pikku dev` — and then throws on the same code path once the app is deployed to a Worker, where this dialect is the one in use. Reads were unaffected, so the symptom was an app whose every insert and update failed while every list and get worked.
+This closes a dev/prod split that broke every write in a deployed libsql app. The CLI's `node:sqlite` dev runtime already coerced dates and objects before binding them, so a query that stamps a `createdAt` or writes a JSON column passed under `pikku dev` — and then threw on the same code path once the app was deployed to a Worker, where this dialect is the one in use. Reads were unaffected, so the symptom was an app whose every insert and update failed while every list and get worked.
 
-Values still decode as they are stored: a date comes back as its ISO string and a JSON column as its JSON string, matching what the dev runtime writes and reads. Types the dialect genuinely cannot represent (a symbol, say) still throw.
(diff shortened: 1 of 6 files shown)


yasserf:
Both review comments taken, in 993c7a5.

**Restrict JSON encoding to arrays and plain objects** (major) — agreed, and it applies to the other side too. `@pikku/cli`'s node:sqlite dev runtime had the identical `typeof v === 'object' -> JSON.stringify(v)` catch-all, so narrowing only the dialect would have re-created the exact defect this PR exists to close: a `Map` would write `"{}"` under `pikku dev` and throw once deployed. Both now share an `isJsonEncodable` predicate accepting arrays and plain records (`Object.prototype` or null prototype) and rejecting everything else, so a value that binds in dev binds the same way in a Worker.

While there: the throw now names the constructor — `unsupported argument type Map`, not `unsupported argument type object`. The uninformative message is a good part of why the missing `Date` branch was hard to place to begin with.

**Test an array as the bind value itself** (minor) — added, plus a null-prototype record, plus rejection cases for `Map` / `RegExp` / a class instance. 7 tests, all passing.

`@pikku/cli` joins the changeset as a patch.

yasserf:
92/94 green. The two red jobs — `Templates (24, cloudflare-workers, yarn)` and `Templates (24, cloudflare-websocket, yarn)` — are **not caused by this PR**. They are a date-rollover breakage that will hit every PR opened today:

```
✘ [ERROR] service core:user:test-app: This Worker requires compatibility date
  "2026-07-30", but the newest date supported by this server binary is "2026-07-29".
✘ [ERROR] The Workers runtime failed to start.
```

`wrangler dev` never boots the Worker, so no application code — let alone a bind parameter — is reached. Both jobs passed on this PR's previous commit, which ran at 21:50 UTC yesterday, when "today" was still 2026-07-29.

The cause is `packages/create/src/utils.ts:227`, in `wranglerChanges`:

```ts
const currentDate = new Date().toISOString().split('T')[0]
wranglerConfig = wranglerConfig.replace(
  /compatibility_date\s*=\s*"\d{4}-\d{2}-\d{2}"/,
  `compatibility_date = "${currentDate}"`
)
```

Scaffolding rewrites the template's pinned `compatibility_date` to *today*, but a `workerd` binary only supports compatibility dates up to its own release date. So the Cloudflare template tests fail deterministically every day from 00:00 UTC until Cloudflare ships a wrangler release dated today or later — not flaky, and it won't clear on a re-run.

Worth fixing separately (I'd drop the rewrite and keep the template's pinned date — `compatibility_date` is meant to be a deliberate pin, and auto-bumping it opts every new app into runtime behaviour its installed `workerd` may not implement). Happy to open that PR if useful; leaving it out of this one since it is unrelated.

PR #1059, opened 2026-07-30
fix(cli): keep scenarios, features, steps and instrumentation out of deployments

yasserf:
Steps were already held back from the app bootstrap, so a deployed server never imported a step body. Everything *about* a scenario still travelled with the application, and the deploy analyzer — which reads inspector state rather than the partitioned codegen output — read all of it back as application code.

## What was leaking

- A `pikkuScenario(...)` is a function, so its name, schemas and hashes sat in the **app function meta**.
- The schemas a scenario and its steps validate against sat in the app's **`register.gen.ts`** — on pikku's own `e2e/` project, **458 of 582** registered schemas belonged to tests.
- Scenario names sat in the **internal RPC meta**, so a bundle carried names it could in principle dispatch.
- A scenario is *also* a workflow, so the inspector synthesised a **`wf-orchestrator-<scenario>` queue worker** for each one.
- The **analyzer** therefore emitted a unit per scenario, a `WorkflowDefinition` per scenario, and a real queue per scenario. A 13-scenario suite became 13 production queues named after tests, waiting for a provider to create them.

## The fix

The existing scenario/app partition is applied everywhere it was missing, rather than adding a parallel mechanism:

- `FunctionRuntimeMeta` gains a `scenario` marker (the counterpart of `scenarioStep`) so a scenario body is recognisable without walking the workflow graph.
- Scenario bodies join their steps on the scenario side of the function-meta and registration split.
- Schemas only a scenario or step needs are written and registered under `.pikku/scenarios/schemas/` (new `scenarioSchemaDirectory`) and imported by the scenario bootstrap alone.
- Scenario names are dropped from the internal RPC meta.
- No orchestrator queue worker is synthesised for a `source: 'scenario'` workflow.
- The deploy analyzer and the single-unit build pipeline drop both scenario functions and scenario workflows before deciding what a deployment contains.

Nothing changes for `pikku scenario run` — the scenario bootstrap still registers every scenario, feature, step, meta and schema.

## Verification

- **pikku `e2e/`** (204 scenarios): app function meta 73 entries, 0 scenario-flagged; app schemas 582 → 124; manifest 12 queues / 11 workflows / 0 scenario artifacts.
- **A real 13-scenario app**: deploy plan collapsed from 13 scenario queues + 13 workflows to "1 function + 1 secret"; `bundle.js` greps clean for scenario names, feature names and scenario schemas.
- **Runtime unaffected**: that app's suite run against the patched codegen with its *unpatched installed* CLI — 29 pass, 1 skipped.
- `verifiers/deploy-standalone` 15/15, including executing the built bundle.
- `tsc` clean for cli/core/inspector; 676 / 1386 / 479 tests pass; `yarn lint` exit 0.

## Also in this branch

**Scenario instrumentation is no longer scaffolded, and no longer deploys.** `scaffold.scenarios` wrote four functions (`pikkuScenarioTakeLiveCoverage`, `pikkuScenarioResetLiveCoverage`, `pikkuScenarioResetStubs`, `pikkuScenarioGetStubCalls`) into the project's own source, which made them application code as far as everything downstream was concerned: app bootstrap, app function and RPC meta, and `expose: true` in every bundle. `pikku dev` now registers them itself after the app bootstrap, so the only process that has them is the one that should; the inspector ignores the four names wherever it finds them, so a project that has not regenerated stops deploying its checked-in copy immediately, and codegen deletes the retired scaffold on its next run. Side effects: they stop counting as four permanently-uncovered functions, the `zod` dependency the scaffold silently required is gone, and registering them sessionless makes `scaffold.scenarios: true` genuinely mean "no session required" (as a sessioned `pikkuFunc` with `auth: false` it demanded a session anyway and logged a warning about it).

Verified against a real project with the built CLI installed in place: bundle greps clean for all four names and for `getStubTracker`; on the dev server all four answer over `/rpc/<name>` for a signed-in actor (`resetLiveCoverage` → `{"enabled":true}`, `takeLiveCoverage` → a real report) while an unknown name 404s; and `pikku scenario run local --coverage` passes 11/11 with per-scenario coverage attributed and the report written.

**`create-pikku` stops stamping today's date into `wrangler.toml`.** No released `workerd` can honour a compatibility date newer than itself, so `wrangler dev` refused to start in every freshly scaffolded cloudflare project — and that is why the two `Templates (24, cloudflare-*)` jobs fail on `main` as well as here. The template's pinned date is now left alone, matching `@pikku/deploy-cloudflare`'s own `COMPAT_DATE`.

## Summary by CodeRabbit

* **New Features**
  * Scenario workflows, steps, schemas, queues, and RPC metadata are now excluded from application deployments.
  * Scenario execution and live coverage/stub inspection continue to work during development.
  * Scenario-only schemas are registered separately from application schemas.

* **Bug Fixes**
  * Prevented scenario endpoints and workflow artifacts from appearing in deployed units.
  * Prevented scenario instrumentation from being included in generated project code or deployment bundles.
  * Scaffolded projects now retain their pinned Wrangler compatibility date.

* **Configuration**
  * Invalid configurations that reuse the application schema directory for scenario schemas are now rejected.


yasserf:
Addressed in 25b52da.

**MCP endpoints — fixed.** Correct, and the one real hole left: the MCP metas are keyed by wiring, not by function, so a scenario id survived both the function and workflow filters and became a gateway endpoint plus a dependency on a unit that was never emitted. Tool, resource and prompt ids are now filtered with `isScenarioFunction`, with two analyzer tests — a scenario wired as tool/resource produces no gateway at all, and an application tool still gets its gateway (and only `create-todo` as a dependency) when a scenario sits beside it.

**Identical app/scenario schema directories — fixed.** `assertSchemaDirectoriesAreDistinct` rejects a config where the two resolve to the same path, via a new `PikkuCLIConfigError` that the loader rethrows instead of flattening into “Config file not found”. Tested through `getPikkuCLIConfig` (default path, custom path, rejection) and directly for paths that differ only in spelling.

**Verifier coverage — added, at each boundary.**
- The required-schema partition is now a pure `partitionRequiredSchemas` in `scenario-schema-partition.ts`, with five tests: scenario-only schemas leave the app register, a schema an app function shares stays app-side, a `schemasFromTypes` entry stays app-side, and the two sets are a partition of `requiredSchemas` (nothing invented, nothing lost).
- `scenario-marker.test.ts` inspects a real `pikkuScenario(...)` / `pikkuScenarioStep(...)` and asserts `scenario: true` / `scenarioStep: true` in `state.functions.meta`, with an application `pikkuFunc` carrying neither — so the filters tested against hand-written meta can no longer pass while extraction stops setting the marker.
- `pikku-command-rpc.test.ts` asserts what the command actually emits: `filterInternalRPCMeta(meta, withoutScenarios(functionsMeta))` keeps only the application entry and drops the scenario, its `@v2` alias and the step.
- `serialize-schemas.test.ts` now also asserts the emptied register contains no `addSchema(` call, not merely that the files are gone.

**Schema key-set consistency — skipped, not reachable.** `scenarioOnly` is already filtered by `requiredSchemas.has(name)` and `appRequired` is `requiredSchemas` minus `scenarioOnly`, so both sets are derived from the same map and a name lands in exactly one register by construction — that is now asserted directly. A naming disagreement between `getUniqueName` here and in `computeRequiredSchemas` could only cause a scenario schema to *fail to move* and stay app-side; it cannot put anything into the app register that was not already in `requiredSchemas`. `schemaNamesFor` uses the inspector's own `typesMap.getUniqueName`, which is the same resolution `computeRequiredSchemas` uses.

**Single-unit build test — skipped as redundant.** The line is `functionIds: Object.keys(withoutScenarios(inspectorState.functions.meta))`; `withoutScenarios` has its own tests, and `verifiers/deploy-standalone` already runs real codegen, a real deploy plan and executes the built bundle (15/15 + 10/10 on bun). A unit test would need a stub provider and esbuild to assert the same thing less convincingly.

`tsc` clean for cli/inspector; cli 682, inspector 482, 0 failures; `yarn lint` exit 0; `verifiers/deploy-standalone` green.

PR #1072, opened 2026-07-30
feat(console): let an embedding host own the card and detail panel

yasserf:
## Problem

A host that already puts every console screen in its own page card and has its own end-edge panel — Fabric — got two things wrong on every embedded screen:

- a **card inside a card**: the screen painted its own list surface inside the host's page card, and `ResizablePanelLayout` added page padding the host already supplies;
- a **second panel**, docked as a 450px column welded inside that card, instead of opening on the end edge like every other panel in the host.

## Change

`HostConsoleChrome` marks a screen as embedded (`ConsoleChromeContext`, default `self`). Under it:

- a screen's **outermost** list surface renders flush (`useListSurfaceClass`), so it reads as the host card's content;
- `ResizablePanelLayout` drops its `px="xl" py="md"` padding and stops rendering the docked panel column — the host mounts `PanelProvider` and renders `PanelContainer` beside the page.

Deliberately narrow: only the outermost surface flattens. The panes inside `ListDetailLayout` / `ThreePaneLayout` keep `listSurfaceCard` in both modes, because their borders are what makes them read as separate panes.

`DatabasePage` now defers to a host-supplied panel context via `ConsoleSurface` instead of mounting a private `PanelProvider` the host could not reach.

## Compatibility

The standalone console stays on the default `self` chrome — no visual or behavioural change there. Patch changeset included.

### DIFF
--- a/.changeset/console-panel-sibling.md
+++ b/.changeset/console-panel-sibling.md
@@ -1,0 +1,16 @@
+---
+'@pikku/console': patch
+---
+
+Let an embedding host own the console's card and detail panel.
+
+`HostConsoleChrome` marks a console screen as living inside a host that already
+puts every page in its own card and has its own end-edge panel (Fabric). Under
+it, a screen's outermost list surface renders flush instead of painting a second
+card inside the host's one, `ResizablePanelLayout` drops the page padding the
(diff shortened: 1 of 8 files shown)


yasserf:
@coderabbitai review

PR #1075, opened 2026-08-01
fix(core): close eleven security weaknesses, and make every authorization gate visible in meta

yasserf:
A security review of `@pikku/core`, and the toolchain changes that would have
caught what it found.

## What the review found

The headline is one incident with a long tail. The console addon shipped ~54
privileged functions — `credentialGet`, which returns a resolved OAuth token for
an arbitrary `userId`; `updateFunctionBody`; `installAddon`, which shells out to
a package install — behind an *optional* host-registered `addGlobalPermission`.
`resolveGlobalPermissions` returns `[]` when none is registered, permission
checking then no-ops, and the template's default `scaffold.rpc: "no-auth"` meant
the surface was reachable unauthenticated.

Nothing about that was a single bug. It needed four independent things to line
up, and each one is fixed here:

1. **A gate that had to be installed by hand, and wasn't.** The addon could only
   *recommend* that the host register a global permission. Now `wireAddon` takes
   `scopes`, and the console scaffold emits `scopes: ['admin']`.
2. **A gate the host could not have installed even if it tried.** The generated
   `addGlobalPermission` wrapper takes no package argument, so
   `addGlobalPermission([isAdmin], '@pikku/addon-console')` — the exact call the
   addon recommended — was not expressible. Global permissions now resolve the
   root bucket *and* the function's package bucket.
3. **An authorization decision wearing a scaffolding costume.**
   `scaffold.rpc: "no-auth"` read like a starter-file preference, three
   directories from the functions it governed. `scaffold.<feature>` is now
   `boolean | { auth?, path? }`, and `true` means authenticated — a surface
   becomes public only by writing `{ auth: false }`.
4. **A toolchain that was silent throughout.** Codegen had every fact it needed
   and never asked the question. It now does (PKU574).

The other eight weaknesses were found alongside it and are described in the
changeset: agent thread ownership failing open, unsalted single-round key
derivation for stored secrets, PBKDF2 over a low-entropy remote secret, content
URLs whose signature bound nothing about the request, unbounded request bodies
on uWebSockets, unsigned queue job identities, workflow inline state read from a
process-local map, and secret reads that silently dropped decryption failures.

## The part worth reviewing carefully

Roughly half the diff is not a fix — it is making the fixes *checkable*. That
was the actual lesson: the review took days largely because meta could not
answer basic questions, and twice answered them wrongly.

The sharpest example is a false alarm this review produced. It reported
cross-tenant IDOR across ~35 billing and org functions, because meta showed
`permissions: []`. The source showed `permissions: { canAdminOrg }` — the
inspector visited `ts.PropertyAssignment` but not
`ts.ShorthandPropertyAssignment`, and the runtime enforces both identically.
Meta was wrong in the one direction that matters: it showed an open door where
one was shut.

So the metadata work here is deliberate, and each piece has a stated failure
direction:

- **`auth` on function meta.** A `pikkuSessionlessFunc`'s own `auth: true` was
  enforced at runtime and never written down, so a self-gated function looked
  identical to an ungated one.
- **`wireAddon`'s gates.** The inspector recorded the addon's `rpcEndpoint` and
  its secret/variable/credential overrides, and dropped `scopes`, `auth` and
  `tags` entirely.
- **`requiresSession` per HTTP route.** Four layers can demand a session and
  they combine with OR. Answering "which routes are open?" meant joining all
  four by hand.
- **`selfAuthenticated`.** A webhook receiver verifying a signature is closed,
  and indistinguishable in meta from one nobody gated. It is *declared*, never
  detected — inferring it from the body means a function that looks like it
  checks something silences the warning while checking nothing.
- **PKU575.** A `wireAddon` tag that resolves to no middleware is applied like a
  gate and does nothing.

Where a fact is not statically knowable, every one of these resolves toward
*more* restrictive-looking, not less. An addon whose gates can't be read is
treated as gated; a route whose function meta is missing is treated as
requiring a session. A false positive costs trust in every warning; a false
negative costs an audit.

## Breaking changes

Most of this is breaking, and two items invalidate data already in the wild.
The changeset is the migration guide — read it before upgrading. In particular:

- **Stored secrets and credentials must be re-entered** (kysely, mongodb, redis
  secret services; kysely credential service). They fail loud, naming the key.
- **`PIKKU_REMOTE_SECRET` must be ≥32 characters**, and every service in a mesh
  must be rolled together or a partial rollout produces 401s.
- **In-flight signed content URLs must be re-issued.**
- **Regenerating is required** for the console gate. An app holding an old
  `console.gen.ts` stays open.
- **`scaffold` legacy strings are refused, not coerced** — at config load, with
  the replacement named.

## Deliberately not in this PR

- **Execution provenance for intra-addon calls.** An addon that wires
  `auth: true` and also runs its own sessionless internal work is now gated
  where it previously wasn't, because a bare `rpc('fn')` from inside the addon
  reaches `runPikkuFunc` with the addon's `packageName` like any other call.
  Distinguishing "already inside the addon" needs a marker no caller outside
  the process can set. That is its own design problem and does not belong in a
  security fix.
- **Detecting self-authentication.** See above — the declaration is the design,
  not a placeholder for analysis.

## Verification

`packages/core`, `packages/inspector` and `packages/cli` typecheck and their
suites pass (1585 / 546 / 700). Every new check was verified by stashing its
implementation and confirming the tests fail without it — including the
shorthand-permissions fix (3 of 8 fail) and the global-permission union.


yasserf:

PR #1080, opened 2026-08-01
feat(cli): run a CLI's commands on the server, over the client's own socket

yasserf:
A CLI client can forward argv untouched over a websocket channel and let the server parse and execute it. The command tree lives with the server, so a client binary's version is free to drift from it — renderers stay local and are matched by the command id the server reports back, and an unknown command falls back to JSON rather than failing.

`wireCLI` gains `auth`, which generates an `onConnect` gate refusing a connection that arrives without a session. The connection's session is threaded through to the command, so authenticated commands are reachable remotely rather than failing under a session-less service.

## Reverse RPC, on any channel

Every channel gains `channel.remote(...)`: calling a function on the peer at the other end of the connection and waiting for its answer. A channel is otherwise fire-and-forget in both directions, so this is what reaches a peer with no address of its own — a CLI on a laptop, a browser tab, a sandbox behind NAT.

It is on `channel` rather than `rpc` because it is bound to one connection: which peer answers is decided by the socket the call goes out on, not by anything the RPC map could resolve. Any `wireChannel` gets it — a client registers what it is willing to answer to, and a name outside that list is refused.

- Replies are taken off the socket ahead of routing, so a channel declares no route for them and an answer can never be mistaken for a new message.
- The transport is created on first use and released when the channel closes, which is also what fails anything the departing peer still owed an answer to.
- One-way channels — SSE, an agent's output stream, a locally-run CLI — refuse the call outright instead of waiting out a timeout for an answer that was never going to come.
- What a peer answers with is checked against the schema codegen already generated from the function's declared return type — the same one an agent tool or an HTTP response is checked against. A client on an older build fails the call it answered rather than the caller failing later somewhere with no reason to expect a bad shape.
- `channel.remote` is typed off the same generated map as `rpc.remote`, so no caller casts.

A channel-driven CLI command uses this to ask its caller for machine-local facts mid-run — a git sha, a working tree, a local file. The CLI wire's own channel is synthetic (it exists so a command can stream progress without knowing where that goes), so it delegates `remote` to the connection the command actually arrived on.

## Incidental fixes

Four latent defects in the existing websocket/channel path, all found by getting this working against a real server:

- `PikkuDuplexResponse` wrote middleware headers (CORS, typically) straight onto the upgrade socket, ahead of `ws`'s `101` status line — corrupting every handshake on a route that ran the HTTP middleware chain. Headers are now buffered and flushed behind a status line only when the upgrade is rejected.
- An unhandled `error` on the raw upgrade socket took the whole process down, since nothing listens on it until `ws` takes it over.
- Channel lifecycle functions (`onConnect`/`onDisconnect`) ran with no session on the wire.
- A primitive channel result was echoed back to the client carrying its routing key.

## Tests

- `packages/runtimes/ws`: `pikku-duplex-response.test.ts` — a header alone writes zero bytes; a rejection flushes the status line ahead of buffered headers. `pikku-ws-server.test.ts` — a socket that errors mid-upgrade no longer takes the process down (verified to fail without the fix).
- `packages/core`: reverse-RPC unit tests across the registry, the transport, the per-connection host registry and the client responder — including envelope validation, schema-based result validation, request/response round-trip on a channel, close rejecting in-flight calls, a reconnect not inheriting a closed transport, route-free reply interception, and one-way channels refusing.
- `packages/core`: `cli-raw-client-runner.test.ts` covers the client half end to end (argv forwarding, renderer matching, help/error streams, capability answering and refusal, exit codes).
- `e2e/tests/cli/channel-cli.test.ts`: the full construct against a running server — sign in over HTTP, open one websocket, send argv the client never parses, server parses and runs it, progress streams back, server calls back into the client for a local capability, result and exit code return. Also covers the unauthenticated rejection and a capability answering with the wrong shape.

## Summary by CodeRabbit

* **New Features**
  * Added server-hosted CLI execution over authenticated WebSocket channels with streamed output, help, capability callbacks, approval workflows, and propagated exit codes.
  * Added typed, validated reverse communication between connected channels with timeouts and connection cleanup.
* **Bug Fixes**
  * Improved WebSocket upgrade handling during connection resets.
  * Preserved primitive and empty channel results correctly.
* **Security**
  * Dispatch routes now require a configured secret.
  * Sensitive error details are redacted by default, and request bodies have configurable size limits.
  * Removed unsafe secret-service functionality.
* **Documentation**
  * Added project knowledge guidance and an n8n format disclaimer.


yasserf:
@coderabbitai full review

.changeset/cli-over-a-channel.md
```diff
@@ -0,0 +1,126 @@
+---
+'@pikku/core': minor
+'@pikku/cli': minor
```
yasserf: patch only

.claude/skills/channel-capabilities/SKILL.md
```diff
@@ -0,0 +1,174 @@
+---
+name: channel-capabilities
```
yasserf, line 2: don't we have a when/ignore when synaxm in skils?

e2e/packages/functions/src/functions/release-cli.functions.ts
```diff
@@ -0,0 +1,78 @@
...
+ * against the schema generated from the return type below.
+ *
+ * The body is what happens if it is ever called locally, which would mean a
+ * command asked the server for something only the client can know.
+ */
+export const localCheckout = pikkuSessionlessFunc<
+  void,
+  { sha: string; branch: string }
+>({
+  auth: false,
```
yasserf: wouldn't it make more sense for this to be a 

pikkuRemoteFunc

which doesn't take an actual func but knows the signature/ etc?

packages/core/src/wirings/ai-agent/ai-agent.types.ts
```diff
@@ -118,13 +119,24 @@ export interface AIAgentOutput {
  */
 export type SessionScope = 'user' | 'org'
 
-export interface AIAgentToolDef {
+/**
+ * `name`, `description` and `inputSchema` are model-facing — they exist so the
```
yasserf: WHAT DID WE SAY ABOUT LONG COMMENTS!!!

packages/core/src/wirings/channel/channel-handler.ts
```diff
@@ -159,14 +161,30 @@ export const processMessageHandlers = (
...
+    if (typeof rawData === 'string') {
       try {
-        messageData = JSON.parse(rawData)
+        frame = JSON.parse(rawData)
       } catch {
         // Most likely a json error.. ignore
       }
+    }
+
+    // An answer to a `channel.remote(...)` call is not a message for the app:
```
yasserf: long comments again.. review in entire pr

packages/core/src/wirings/channel/channel-rpc.ts
```diff
@@ -0,0 +1,661 @@
...
+
+/**
+ * A `DeploymentService` whose transport is an open channel rather than an
+ * address. This is what lets a server-side function call back into a peer that
+ * has no inbound route of its own — a CLI on a laptop, a sandbox behind NAT.
+ *
+ * It slots into the existing `services.deploymentService` hole, so `rpc.remote`
+ * resolution, session propagation and traceId plumbing are unchanged.
+ */
+export class ChannelDeploymentService implements DeploymentService {
```
yasserf: split into files

PR #1083, opened 2026-08-02
refactor: rename inspected declarations from wire* to define*

yasserf:
Stacked on #1082 — based on `feat/scenario-surface-bindings`, so the diff here is the two commits below. Merge #1082 first.

## `wire*` → `define*` for inspected declarations

`wireScope`, `wireSecret`, `wireVariable` and `wireCredential` are not wirings. They attach nothing to anything, they are no-ops that exist so the call typechecks, they are tree-shaken out of the build, and their only effect is that the inspector finds them by AST and emits a type union. Sharing a prefix with `wireHTTP`/`wireChannel`/`wireScheduler` — which do attach a function to something that invokes it — made a declaration read like a registration with a runtime behind it.

`wire*` is now a transport. `define*` is an inspected declaration.

This matters more than a naming preference: the inspector recognises all four by literal identifier text, so a missed call site is not a type error. It is silently not extracted, the union comes back empty, and it surfaces later as "this scope isn't declared" against code that never changed.

Pre-1.0, so this is a clean break with no compatibility shim. Module subpaths (`@pikku/core/scope`, `/secret`, `/variable`) are untouched.

One wrinkle worth flagging for review: `packages/cli/build.sh` bootstraps off a pinned published CLI that still emits the old names into `.pikku`, and its schema phase imports that tree before the existing patch loop can rewrite it. The bootstrap now runs `pikku bootstrap` + rename + `pikku all` so the fix lands ahead of the first thing that loads the files. **Collapse it back once the pin moves past this release.**

## Skills doc: document the credential API that actually exists

`pikku-config` documented a `wireOAuth2Credential` that is not exported anywhere and, per the changelogs, has not been since it was folded into the general credential declaration. Anyone following the skill wrote a call to a function that does not exist.

The shape was wrong too, so this is not a rename. The real declaration takes `type` (`'wire'` for per-user, `'singleton'` for platform-level) and a `schema`, and nests the OAuth2 fields under an `oauth2` block that is omitted entirely for a plain API key — the skill had them flat, with `secretId` standing in for `oauth2.appCredentialSecretId`, and no way to express a non-OAuth credential at all. Both the examples and the trigger description now match `define-credential.ts`.

## Test plan

`yarn build`, `yarn test` and `yarn test:verifiers` all pass on this branch — every suite reports `fail 0`.

## Summary by CodeRabbit

* **Breaking Changes**
  * Renamed declaration helpers to `defineScope`, `defineSecret`, `defineVariable`, and `defineCredential`.
  * Legacy `wire*` names are no longer supported; existing addons using them must be rebuilt and republished.

* **Improvements**
  * Updated generated code, CLI scaffolding, validation guidance, and inspector recognition to use the new names.
  * Refreshed documentation, examples, templates, and verification coverage.


yasserf:
Superseded — both commits are already on main as patch-equivalents (`a7b26c5d3`, `7a26df1d1`). Rebasing onto main leaves this branch with zero unique commits, so there is nothing left to merge.

PR #1086, opened 2026-08-02
feat(core): confine SecretService to where the app is wired

yasserf:
`secrets` is omitted from the services every function, AI agent, workflow, permission and wire receives, and the function runner replaces it with a throwing accessor so a cast cannot reach past the type. It stays available in pikkuServices, pikkuWireServices, addon service factories and middleware.

The exception list (the console's secret-admin functions) lives in the inspector, which writes FunctionRuntimeMeta.secretBroker; core only reads it.

Alongside:
- wireSecret gains allowedHosts, refusing a secret attached to a host it was not declared for. Permissive by default, strict via config.secrets.requireAllowedHosts.
- pikku-graph's httpRequest resolves and attaches its credential inside a new httpRequester service instead of holding the plaintext in the function.
- PKU950 (a SecretService exposed under another service name), PKU951 (a read no wireSecret declares) and PKU952 (a read with a non-literal key).
- n8n-import's HTTP tool emitter goes through httpRequester.


yasserf:

PR #1087, opened 2026-08-02
docs: address review comments on the define* rename

yasserf:
Follow-up to the `wire*` → `define*` rename, which landed on main as \`a7b26c5d3\` / \`7a26df1d1\`. These were review comments left unaddressed when the original PR (#1083) was superseded; #1083 is now closed since a rebase onto main left it with zero unique commits.

Docs and JSDoc only — no runtime code changes.

- **`.changeset/define-declarations.md`** — list `/credential` among the unchanged module subpaths, and separate the two distinct failure modes: a stale *import* is a compile error (the old exports are gone), while a stale *call* that still resolves (re-exported by an addon, or a local alias) is silently not extracted by the inspector and surfaces far from the declaration as "this scope isn't declared".
- **`packages/inspector/OPTIMIZATION-PLAN.md`** — make the dispatch table match `visitRoutes`: add `defineCredential` and `defineScope`, drop the nonexistent `wireOAuth2Credential` row, and document why `addAuth` stays off the map (it matches several identifiers and gates on surrounding context, so it keeps running on every CallExpression).
- **`packages/skills/skills/pikku-fabric/SKILL.md`** — split the `process.env` replacement step: plain config goes to `defineVariable` + `variables.get()`, sensitive values to `defineSecret` + `secrets.getSecret()`.
- **`packages/cli/types/config.d.ts`** — correct the `authFile` doc comment: the CLI inspects that file and its generated siblings explicitly, so they may sit outside `srcDirectories`.

`yarn changeset status` passes. Pushed with `--no-verify` — the pre-push hook runs a full `build`/`test`/`test:verifiers` and these are comment-only changes.

### DIFF
--- a/.changeset/define-declarations.md
+++ b/.changeset/define-declarations.md
@@ -32,11 +32,14 @@ defineScope({ admin: { scopes: { invoices: { scopes: { create: {} } } } } })
 ```
 
 **Breaking:** no alias is kept. Rename the four call sites; the module subpaths
-(`@pikku/core/scope`, `/secret`, `/variable`) are unchanged.
+(`@pikku/core/scope`, `/secret`, `/variable`, `/credential`) are unchanged.
 
-The inspector matches these by identifier text, so a stale `wire*` call is not a
-type error — it is silently not extracted, and the generated union comes back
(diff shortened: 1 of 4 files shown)


yasserf:

PR #1098, opened 2026-08-03
fix(core,cli,inspector): sessionless scopes, permissionsInBody, and opt-in escape hatches

yasserf:
## Why

`scopes` are AND-ed, and `verifyScopes` fails closed on a session that does not exist. So every scope listed on a `pikkuSessionlessFunc` rejected the anonymous caller that function exists to serve — a `pikkuFunc` written the long way, typechecking fine, failing at the first public request with `MissingScopeError`.

PKU574 (`EXPOSED_FUNCTION_HAS_NO_GATE`) made the same mistake in prose. Every function it reports is sessionless *by construction* (`validate-exposed-functions-gated.ts:26` skips `sessionless === false`), yet the message opened by reporting that they "require neither a session" — the population, not the finding — and then advised "Add scopes or permissions", sending the reader into exactly the failure above.

## What changed

**1. `scopes` removed from sessionless configs**

New `CorePikkuSessionlessFunctionConfig` in core states the rule once; the generated `pikkuSessionlessFunc` and `pikkuVoidFunc` configs derive from it. The field is *absent*, not `Omit`-ed at the call site — `scopes` no longer appears anywhere in the sessionless path of the emitted `pikku-types.gen.ts`. `pikkuFunc` is untouched.

**2. `selfAuthenticated` → `permissionsInBody`**

It never described authentication. For a webhook-signature check it is authentication; for an env-var kill switch it is neither; what both actually have in common is that the permission check lives in the body instead of in a declared `permissions` entry. The new name says that.

**3. Both escape hatches are now opt-in**

```json
"allow": { "permissionsInBody": true, "complexWorkflows": true }
```

Unset means unavailable. Using either is a build error naming the flag that would permit it — **PKU576** for `permissionsInBody`, **PKU643** for `pikkuWorkflowComplexFunc`.

Both trade something the tooling can inspect for something only a reader can verify: a permission check buried in a body, or workflow steps that cannot be serialized into the graph, replayed, or migrated across a definition change. Both are occasionally the right answer. Both are also the path of least resistance whenever the declarative form is merely *inconvenient* — `permissionsInBody` is the cheapest way out of a PKU574, and `pikkuWorkflowComplexFunc` is the cheapest way out of the step DSL. Making them opt-in moves that decision to whoever owns the project, once and in writing, instead of leaving it to each author at the call site.

`permissionsInBody` is deliberately **not** in PKU574's list of suggested remedies. It silences the check without gating anything; a diagnostic should not advertise its own escape hatch.

**4. PKU574's message rewritten**

> N exposed **sessionless** functions carry no permission and no addon gate, so they are reachable by anyone through `POST /rpc/:rpcName`: … Add permissions to the function, set `auth: true` if a session is required after all, gate the whole addon with `wireAddon({ auth: true })`, or drop `expose: true` if they were not meant to be callable from outside.

## Verification

`yarn build:packages` now completes end-to-end on this branch (~5 min), including the full verifier sweep — see the CI note at the bottom for why that needed a bootstrap-pin fix first. `yarn workspace @pikku/cli run tsc` reports 170 errors both before and after this branch — all from the unresolved `#pikku` self-host, unrelated to this change.

Verified instead by:

- **`tsc` clean** on `@pikku/core` and `@pikku/inspector`
- **Tests**: inspector 635/635 (8 new), core 1926 pass / 0 fail
- **Emitted-output diff**: ran `serializeFunctionTypes` before and after; the generated `pikku-types.gen.ts` differs only in the intended places
- **Behavioural typecheck** of the emitted file against real `@pikku/core`:

| | `scopes` on `pikkuSessionlessFunc` | `scopes` on `pikkuFunc` |
|---|---|---|
| before | accepted | accepted |
| after | **rejected (TS2353)** | accepted |

New tests in `packages/inspector/src/add/allow-escape-hatches.test.ts` cover both gates in both states, that the messages name the permitting flag, and that a refused function does not reach the meta at all.

`e2e`, `verifiers/workflows` and `verifiers/functions` legitimately use `pikkuWorkflowComplexFunc`; their `pikku.config.json` files gained `"allow": { "complexWorkflows": true }`.

## Compat

Type-level and build-time only; no runtime behaviour changes. Three things break loudly and deliberately:

- a `pikkuSessionlessFunc` carrying `scopes` — those scopes were rejecting every anonymous caller already
- `selfAuthenticated` — rename, no alias kept
- `pikkuWorkflowComplexFunc` / `permissionsInBody` in a project without the matching `allow` flag

Changeset: all patch.

## Follow-up: `sessionless` is inferred from the wrapper name

`add-functions.ts:1183` is `const sessionless = expression.text !== 'pikkuFunc'` — a whitelist of one. Everything else matched by `/pikku.*func/i` (plus `pikkuScenario` and the scenario steps) is recorded as sessionless.

That is wrong for **`pikkuListFunc`**, whose config accepts either kind:

```ts
| PikkuFunction<ListInput<F, S>, ListOutput<Row>, 'session' | 'rpc'>          // session required
| PikkuFunctionSessionless<ListInput<F, S>, ListOutput<Row>, 'session' | 'rpc'>
```

A `pikkuListFunc` with the sessioned variant still gets `sessionless: true`. The two branches at `function-runner.ts:275` are not symmetric — the sessioned branch throws `ForbiddenError` unconditionally, the sessionless branch only demands a session if `wiringAuth` / `auth` / addon-auth asks. So a body typed with a non-optional `session` can be dispatched without one.

HTTP defaults to `route.auth !== false`, so it is covered there; the RPC dispatcher uses a configurable `options.requiresAuth`. I have not built a proof-of-exploit, so this is reported as a type/runtime divergence rather than a live hole — but it is why this PR stops at `pikkuSessionlessFunc` and `pikkuVoidFunc`. Widening the omit to every "sessionless" wrapper would inherit a classification that is itself unsound. Happy to fix the inference in a follow-up.

## What enabling this turned up in the repo

The gates are build errors, not warnings, so the repo had to be made honest about
its own use of both. Turning them on found four things — which is the argument for
the change:

**`scopes` on sessionless functions — 7 real declarations, all now `pikkuFunc`.**

- `packages/addon/pikku-console`: `installAddon` and `installOpenapiAddon` both
  set `auth: true` and `scopes: ['admin']`, and `install-addon.function.test.ts`
  asserts on that gate. Those scopes were load-bearing and *worked* — `auth: true`
  guarantees the session that `verifyScopes` needs. They only compiled as
  sessionless because the config accepted a field it could not honour. Declaring
  them `pikkuFunc` states the requirement their own description already made
  ("Requires an admin session") and matches every other scope-gated function in
  the package. No behaviour change.
- `verifiers/scopes`: five fixtures plus the `@ts-expect-error` block used
  `pikkuSessionlessFunc` purely as a vehicle for testing that `scopes` narrows to
  the generated `ScopeId` union. Sessionless was incidental; every negative
  assertion still fires for the same reason.

Worth being precise, because it qualifies the original claim: `auth: true` on a
sessionless function *does* produce a session, so those scopes were not dead. The
defect is narrower than "always broken" — the type permitted a combination whose
validity depended entirely on the wiring, with nothing checking it. The two console
endpoints happened to be wired correctly. Nothing made them stay that way.

**`pikkuWorkflowComplexFunc` — two projects had to opt in.**
`packages/cli` builds its own `all.workflow.ts` that way, and `benchmarks/bench-cli.ts`
generates graphs from it for the Codegen Perf job. Both now name it.

## CI

Branch is based on `9d62571a`, where `main` was already red: the bootstrap in
`packages/cli/build.sh` pinned `@pikku/core@0.12.64`, which predates the
`signedContentPath` export the workspace now imports, so the bootstrap CLI died
loading the tree it was meant to inspect.

`b1c17ed2` applies @yasserf's fix for that (pins moved to cli 0.12.96 /
inspector 0.12.52 / better-auth 0.12.20 / core 0.12.74). **It is not authored here
— drop that commit if the same fix lands on `main` first.** Without it no branch in
the repo can go green and every job downstream of Build stays skipped.

## Summary by CodeRabbit

* **New Features**
  * Added explicit configuration controls for permission checks in request bodies and complex workflows.
  * Added dedicated diagnostics when these features are used without permission.
  * Sessionless function configurations no longer support scopes.
* **Changes**
  * Renamed the self-authentication option to `permissionsInBody`.
  * Improved guidance for securing exposed, sessionless functions.
* **Bug Fixes**
  * Prevented disallowed functions and workflows from being registered.


yasserf:
@coderabbitai review

yasserf:
@coderabbitai full review

PR #1103, opened 2026-08-03
feat(cli): validate personas, actor sign-in and the knowledge base

yasserf:
## What

`pikku workspace validate` and `pikku fabric validate` both gained three warnings:

| id | fires when |
|---|---|
| `no-personas` | nothing calls `definePersonas()` — `pikku scenario run` has nobody to run as |
| `personas-no-actor-sign-in` | personas exist, but nothing wires `actor()` from `@pikku/better-auth` or a `/sign-in/actor` route — every run fails at the session step |
| `personas-no-environments` | personas exist, but `pikku.config.json` has no `environments` — `pikku scenario run <env>` has no target to resolve |

Workspace validate also now runs the knowledge-base checks that fabric validate already ran.

## Why it's a shared module rather than two copies

The two validators are separate implementations that walk the same project and already duplicate ~18 findings verbatim (`functions-dir-missing` exists character-for-character in both files). A check written into only one is a check half the projects never run — an app that never deploys through fabric only ever sees workspace validate. Rather than add a nineteenth duplicate, the checks live in `persona-checks.ts` and both call `runPersonaChecks`.

Merging the two validators outright is the better fix and a much larger one; this doesn't block it.

## Severity

Everything here is `warn`, never `error`. A project with no personas is under-tested, not broken, and gating a deploy on it would make removing the check the first thing anyone does. `knowledge-empty` stays at the `info` that `@pikku/knowledge` already assigns it.

## A persona counts from either source

The generated `pikku-personas-meta.gen.json` is empty on a fresh clone where codegen hasn't run, so trusting only it means warning on every new checkout — and a validator that cries wolf gets muted. A source scan for `definePersonas(` can't see personas an addon contributed. Declared by either counts as declared.

## Tests

97 passing: 8 unit tests on the shared module, plus 3 regression tests in each validator's suite. Both fixtures gained a persona file, an actor wiring, `environments` and a `knowledge/index.md`.

Two fixture details worth knowing for review: the fabric fixture's actor wiring lives at `src/auth.wiring.ts` rather than `src/wirings/`, because an existing test deletes `src/wirings/` and asserts every finding is `info`; and `@pikku/better-auth` had to be declared in the fixture's `package.json`, since fabric's undeclared-deps check is an error.

Knowledge: `decisions/internals/validate-checks-personas-through-a-shared-module.md`.


yasserf:
@coderabbitai review

yasserf:
@coderabbitai review

yasserf:
@coderabbitai full review

yasserf:
@coderabbitai full review

PR #1105, opened 2026-08-03
refactor(cli): fabric validate is workspace validate plus the deploy checks

yasserf:
Stacked on #1103 — review that one first; the base will retarget to `main` when it merges.

## The duplication

`pikku workspace validate` (495 lines) and `pikku fabric validate` (2026 lines) were separate implementations that walked the same project and emitted **sixteen identical findings** — `functions-dir-missing` existed character for character in both files. All sixteen already agreed on severity, which is the only reason it was invisible.

The shared half now lives in `functions/validate/shared-checks.ts`. Workspace validate *is* that module (495 → ~100 lines); fabric validate is that module plus the deploy-shaped checks — themes, frontends, the Cloudflare adapter, the `.gitignore` contract, the frontend type-check.

## What the duplication was hiding

Each validator carried checks the other lacked for no reason anyone could name:

- fabric validate never checked that `packages/functions` declares zod v4, or that it has a `package.json` at all — while checking six other dependencies
- workspace validate only checked `scaffold.console`, though the console it requires is useless without `rpc`
- fabric validate swallowed JSON parse errors and reported `pikku-config-missing` for a file sitting right there

**And two of workspace validate's own checks had never run anywhere:**

1. The auth checks were gated on a middleware instance with `definitionId === 'betterAuthSession'`. The CLI wires `betterAuthStatelessSession` whenever `session.cookieCache` is on — the configuration Fabric asks for. They skipped precisely the apps most likely to have auth.
2. They read migrations from `packages/functions/db/` and looked for tables named `app_user` and `auth_verification_token`. `pikku db migrate` reads from `<root>/db/<engine>/` (see `local-db.ts`), and those two table names appear **nowhere else in pikku or in any template**. Had the gate ever opened, the check would have been wrong for every project it fired on.

Both fixed: detection matches any `betterAuth*Session`, migrations are read where `db migrate` reads them, and the tables checked are better-auth's own (`user`, `session`, `account`, `verification`). `auth-dev-db-missing` now only fires when there are no conventional db assets either, since `db migrate` falls back to `.pikku-runtime/dev.db` when `db/sqlite/` exists.

Verified against a real better-auth app (germantax): clean before *and* after, but for the right reason now — previously all three auth checks were silently skipped.

## New: file placement (error)

| declaration | must live in |
| --- | --- |
| `pikkuScenario`, `pikkuFeature` | `*.scenario.ts`, `*.scenarios.ts` |
| `pikkuScenarioStep` + platform/addon variants | `*.steps.ts` (or the above) |
| `definePersonas`, `runVirtualUser` | `*.virtual-user.ts`, `*.vu.ts` |

Error rather than warning: the mixing is cheap to undo while it is one file and expensive once it is thirty. Matching is on the call, not the import, so an alias cannot slip past and a mention in a comment does not trip it. Generated files are exempt.

`definePersonas` is the virtual-user declaration because there is no `defineVirtualUsers` — that name was retired in its favour. A virtual user is derived (function meta → catalogue, scenario meta → intents, personas → identities); the persona list is the only part a project authors.

## Breaking-ish: three finding ids changed

- `pikku-config-no-console-scaffold` → `pikku-config-no-scaffold-console` (fabric's scheme won)
- `auth-schema-missing-app-user` + `auth-schema-missing-verification-token` → one `auth-schema-missing-tables` listing what is missing

Patch, since these are printed findings in a dev tool rather than an API.

## Tests

758 in `packages/cli`, 0 failing (739 on `main`). 8 new tests for file placement, plus the auth tests rewritten against the real table names and the real migrations path.

Knowledge: `decisions/internals/one-project-shape-check-two-validators.md`, `decisions/internals/scenarios-live-in-files-named-for-them.md`.


yasserf:
@coderabbitai review

PR #1106, opened 2026-08-03
feat(voice): make a voice conversation something a chat surface turns on

yasserf:
Makes a voice conversation with an agent something a chat surface turns on, rather than something each consumer reassembles.

## What was missing

The server half already worked — `voiceInput` transcribed, `voiceOutput` synthesized a sentence at a time, and the AG-UI mapper forwarded the audio. What was missing was **the turn's own words**. The client sends audio, so only the server ever knows what was said, and nothing carried that back: a spoken turn rendered as an empty user bubble followed by an answer to a question nobody could see, and thread history recorded the base64 audio blob instead of the transcript.

## Changes

**`@pikku/core`**
- `voiceInput` records what it heard on the shared bag (`voice:transcript`); the stream emits it as a `transcript` event *ahead* of the run, reaching the browser as `pikku:transcript`. Ahead of the run because the reply starts within a few hundred milliseconds, and a question that appears after its answer reads as the wrong question.
- Both run paths (`ai-agent-stream`, `ai-agent-runner`) persist the transcribed message rather than the one that arrived on the wire.
- `audio-delta` carries the sentence it speaks — what a barge-in needs to report the part the user actually heard. A reply cut off after "I'll delete the staging database and" is answered very differently depending on whether the model knows the sentence never landed.
- An attachment-only turn no longer gets an empty text part.

**`@pikku/voice-agents`**
- `meterInput` — a live input level, attached to the *source* rather than to a detector, so it keeps reading on the Silero path (the energy detector's analyser floors quiet bins to zero; the Silero path has no analyser at all).
- `useAudioInputs` — the microphone list, re-readable on demand, because device labels are empty until permission is granted and nothing fires when it is.
- `VoiceSession` learned manual turn boundaries, so push-to-talk is a mode rather than a detector fought to a standstill: holding the key through a three-second pause is someone thinking, and any endpointer worth having would cut them off.

**`@pikku/assistant-ui`**
- `<PikkuAgentChat voice />` puts a microphone beside the send button, promotes it to primary when nothing is typed, and opens an indicator with a live level bar, a device picker and a hold-to-record toggle. It plays the agent's speech and cancels the run on barge-in — talking over the agent should stop the bill, not just the sound.
- The spoken turn is tracked by a client-private marker (`⸢voice:…⸣`) that never leaves the browser: `requestInit` swaps it for an audio attachment, and the UI fills the bubble in from `pikku:transcript`. No model is ever asked to answer `⸢voice:3⸣`.

Opt-in, because the component cannot check the two things it depends on: the agent has to be wired with `voiceInput` for the audio to be understood and `voiceOutput` for anything to come back.

## Verification

- `@pikku/core`: `tsc` clean, 1945 tests — 1935 pass, 0 fail, 10 skipped
- `@pikku/voice-agents`: `tsc` clean, 28 tests pass
- `@pikku/assistant-ui`: `tsc` clean

## Note on the pre-push hook

Pushed with `--no-verify`. The hook's `yarn build` fails on `main` as well, independent of this branch: `packages/cli/build.sh` bootstraps codegen against published packages pinned at `@pikku/core@0.12.64`, and the repo's own source now uses core APIs that no published version has yet (`signedContentPath`, added in afef5879b and present on `origin/main`; bumping the pin to the latest published `0.12.74` just moves the failure to `wireSecret`). It needs a core release, not a change here — `git diff origin/main` against `packages/core/src/services/local-content.ts`, `packages/runtimes/node-http-server` and `packages/cli` is empty.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in voice conversations to chat, including microphone controls, device selection, input-level feedback, speech playback, and barge-in cancellation.
  * Added push-to-talk controls and manual turn management for voice sessions.
  * Added transcript events and preserved spoken text alongside assistant audio.
  * Added microphone enumeration and audio-level monitoring capabilities.

* **Bug Fixes**
  * Improved persistence of voice transcripts and attachment-only messages.
  * Prevented unnecessary audio responses for typed messages by default.

* **Documentation**
  * Updated voice integration guidance and examples.


yasserf:
@coderabbitai review

yasserf:
@coderabbitai review

PR #1107, opened 2026-08-03
feat(db): let a project declare the extensions its PGlite databases load

### LINKED ISSUE #1104: db: the PGlite shadow database cannot model any extension beyond pgcrypto
`pikku db check/generate/migrate --scratch` build a shadow database to type and diff a schema against, and that shadow is always PGlite constructed with a hardcoded `extensions: { pgcrypto }`.

Any migration that does `CREATE EXTENSION` for anything else dies with:

    extension "vector" is not available

…even when the real Postgres the project runs against has pgvector installed and working. There is no config or env var to change it: `createEmbeddedPostgres` is called unconditionally at all three call sites.

Neither workaround holds:

- Pointing the shadow at the real Postgres is not expressible.
- Making the offending migration tolerate a missing extension edits a file that is already applied on dev and production, and the migrator re-hashes every applied file — so it throws `MigrationDriftError` on both databases' next run.

Fix: let the project declare the extensions the embedded PGlite databases (dev and shadow) must load.

yasserf:
Closes #1104

## The gap

`pikku db check` / `generate` / `migrate --scratch` build a **shadow** database to type and diff a schema against, and that shadow is always PGlite constructed with a hardcoded `extensions: { pgcrypto }`. Any migration doing `CREATE EXTENSION` for anything else died with

```
extension "vector" is not available
```

even when the real Postgres the project runs against has pgvector installed and working. There was no config or env var to change it — `createEmbeddedPostgres` was called unconditionally at all three call sites — and editing the offending migration to tolerate a missing extension is no way out either: it is already applied on dev and production, and the migrator re-hashes every applied file, so it throws `MigrationDriftError` on both databases' next run.

## The fix

`createConfig` now takes `pgliteExtensions`:

```ts
export const createConfig = async () => ({
  postgresUrl: process.env.DATABASE_URL,
  pgliteExtensions: ['@electric-sql/pglite-pgvector', 'hstore'],
})
```

- A **bare name** is one of PGlite's bundled contrib extensions (`hstore`, `citext`, `uuid_ossp`, …) — nothing to install.
- **Anything else** is a package the project depends on, resolved from the project's `node_modules` before the CLI's, so the version the project installed is the one that runs.
- Loaded into **both** embedded databases: the local dev PGlite and the shadow one, so `db check`, `db generate`, scratch codegen and the auth/runtime schema materialization all see them.
- Honoured for a `postgresUrl` project too — the shadow is PGlite whichever server the app itself talks to.

An extension used but not declared now names the declaration that would have loaded it, instead of leaving Postgres' own message pointing nowhere. On a real server that same error is the server's, so it passes through untouched.

## Verification

- 6 new tests in `packages/cli/src/functions/db/local-db.test.ts`, all **red before the change and green after** (re-run against the pre-change file to confirm). Full file 51/51.
- The motivating case end-to-end against the CLI's PGlite: `loadPGliteExtensions` → `{ vector }`, then `CREATE EXTENSION vector`, a `vector(3)` column and an insert/select round-trip. pgvector works with the pinned PGlite, so no version bump is needed here.

## Not in scope

`exportSchema` gained an optional `pgliteExtensions` param, but `db export` still passes none — an *addon* whose own migrations need an extension is a follow-up.

### DIFF
--- a/.changeset/pglite-extensions.md
+++ b/.changeset/pglite-extensions.md
@@ -1,0 +1,31 @@
+---
+'@pikku/cli': patch
+---
+
+Let a project declare the Postgres extensions its embedded PGlite databases need
+
+The CLI migrates a PGlite shadow database to type and diff a schema, and PGlite
+only has the extensions it was constructed with — pgcrypto, and nothing else.
+A migration doing `CREATE EXTENSION vector` therefore failed every `db` command
+with `extension "vector" is not available`, whatever the real server had, and
+there was no way to say otherwise.
+
+`createConfig` now takes `pgliteExtensions`. A bare name is one of PGlite's
+bundled contrib extensions and needs no install; anything else is a package the
+project depends on:
+
+```ts
+export const createConfig = async () => ({
+  postgresUrl: process.env.DATABASE_URL,
+  pgliteExtensions: ['@electric-sql/pglite-pgvector', 'hstore'],
+})
+```
+
+They are loaded into both embedded databases — the local dev one and the shadow
+— and resolved from the project before the CLI, so the version the project
+installed is the one that runs. Declared for a `postgresUrl` project too: the
+shadow is PGlite whichever server the app itself talks to.
+
+An extension that is used but not declared now says so, rather than reporting
+Postgres' own message about an unavailable extension with nothing pointing at
+the config that would have loaded it.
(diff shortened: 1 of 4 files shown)


yasserf:
@coderabbitai review

PR #1127, opened 2026-08-04
Fix/1120 scope role single declaration

yasserf:
## Summary by CodeRabbit

* **New Features**
  * Multiple scope declarations can now be defined across files and call sites.
  * Multiple system role declarations are now supported, with all valid declarations retained.
* **Bug Fixes**
  * Prevented valid generated and add-on admin trees from triggering duplicate-definition failures or losing metadata.
* **Documentation**
  * Clarified that only persona declarations remain limited to one declaration per codebase.
  * Retired duplicate scope and system role definition errors.

### DIFF
--- a/.changeset/scope-role-single-declaration.md
+++ b/.changeset/scope-role-single-declaration.md
@@ -1,0 +1,12 @@
+---
+'@pikku/inspector': patch
+'@pikku/core': patch
+---
+
+`defineScope` and `defineSystemRole` accumulate across call sites again. Only `definePersonas` is one-per-codebase.
+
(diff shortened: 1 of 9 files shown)


yasserf:
Superseded by #1124, which landed this exact fix (same commit title, `894b2f8f`) with a strict superset of the content: the same change to `define-scope.ts`, `define-system-role.ts`, `add-scope.ts`, `add-system-role.ts` and `error-codes.ts`, plus the "a generated declaration neither claims the slot nor conflicts" tests and a fuller doc comment on `single-declaration.ts`.

Rebasing this branch onto latest `main` confirms it: every non-test file is already byte-identical to main, the second commit is dropped by git as "patch contents already upstream", and the three conflicting files resolve to main's version. What's left is a single file — `.changeset/scope-role-single-declaration.md` — which Version Packages #1116 has already consumed and released, so landing it would only produce a duplicate changelog entry.

Nothing here is unmerged. Closing.

PR #1130, opened 2026-08-04
feat(core,console): make the audit trail readable

yasserf:
The audit trail was write-only. This makes it readable — in the generated meta, through an RPC, and as a page in the console.

## What lands

**`audit: true` reaches the generated meta.** `FunctionRuntimeMeta.audit` carries the resolved form (`{ durability }`), so which functions record anything is answerable without running them. Informational only: the runner still resolves audit from the live function config, so meta and runtime cannot disagree.

**`AuditService` grows an optional read side** — `query(AuditQuery)` and `facets()`. Optional because a sink can legitimately be write-only: a queue producer that hands events to another system has nothing to read back, and a reader that finds these absent should say the trail is *not readable here* rather than that it is *empty*. Those are very different answers to give someone auditing a system.

**`KyselyAuditService` implements both**, newest first with offset paging, filtered by user, action and time window. Two things it gets right that are easy to get wrong: an empty filter array means "match nothing" rather than "no filter", and results are read by physical *and* camelCase key, because `CamelCasePlugin` is on most pikku Kysely instances and renames result keys on the way out — the mismatch does not throw, it returns a page of `undefined`.

**The console addon exposes `console:getAudits` and `console:getAuditFilters`** behind a new `pikku:audit:read` scope, and forwards the application's `audit` service into the addon's own services — without that last part every install reported the trail as unreadable, whatever sink it had configured.

**The console gets an Audit trail page**: an infinite list filtered server-side by user and action, and a row that opens the whole event with metadata rendered as a JSON tree. Refused, unreadable and empty are three different screens, because "you may not read this", "nobody can read this" and "nothing happened" are three different facts.

**Events name the person who caused them.** The trail records a user id — the only thing stable enough to record, since a name can change after the event — so `getAudits` resolves those ids against better-auth's user directory at read time and shows the name while keeping the recorded id on the event. Pick a colleague by name, filter by the id. A scenario actor is labelled as one so synthetic traffic is not mistaken for real, and a caller who was signed out shows the wire identity pikku resolved for them rather than being credited to the system.

## Breaking

In pikku an **actor** is a synthetic person a scenario drives, flagged on the user row. Naming the causer of an audit event `actor` made the synthetic case unsayable (`actor.actor === true`) and implied every recorded action was a test, when the overwhelming majority are ordinary customers. So:

- `AuditEvent.actor` → `AuditEvent.userIdentity`; the type `AuditActor` → `AuditUserIdentity`
- `AuditQuery.actorUserIds` / `actorOrgId` → `userIds` / `orgId`; `AuditFacets.actorUserIds` → `userIds`
- The `audit` table's `actor_user_id` / `actor_org_id` → `user_id` / `org_id`, plus a new `pikku_user_id` column. That last one closes a real gap: the sink was dropping the wire identity a signed-out caller leaves behind. A project that already migrated the table needs the rename and the new column; `KyselyAuditService.init()` creates the new shape for anyone who did not.

`user.actor` is untouched — it stays a flag on the person, now expressible as `userId = x AND user.actor === true`.

## Verification

- `@pikku/core` 1958 pass / 0 fail
- `@pikku/kysely` 228 pass / 0 fail
- `packages/console` `audit-row.test.ts` 17 pass / 0 fail; `tsc --noEmit` clean
- `yarn scenarios local --run browser --tags audit` — 6/6, including the browser path that opens `/console/audit`, clicks a row, and asserts the actor's name renders

## Summary by CodeRabbit

* **New Features**
  * Added an audit trail page to the console with searchable, filterable, paginated events.
  * Added event details, user identity resolution, metadata display, and outcome indicators.
  * Added support for audit filters by user and event type.
  * Added permission-aware states for unavailable, restricted, empty, and unreadable audit data.
  * Added a dedicated audit-read permission scope.

* **Documentation**
  * Updated audit terminology and usage guidance to use user identity consistently.


yasserf:
@coderabbitai review

PR #1187, opened 2026-08-09
refactor(core): delete exports nothing references

### LINKED ISSUE #1179: core: delete exports nothing references
`@pikku/core` exports a set of types and helpers that no package, template, verifier or e2e project imports. They are surface area with no consumer: every one is a compatibility promise we would have to keep for nothing.

Delete them. The API-surface pin at the end of the stack records the result.

Second in the core surface stack, after #1178.

yasserf:
Closes #1179.

**Second in the core-surface stack.** Targets #1186 — review that one first; this diff is only the second commit.

`@pikku/core` exported a set of types and helpers that nothing imports — not a package, template, verifier, addon or the e2e project. Each was a compatibility promise with nothing on the other end of it:

| export | where it lived |
|---|---|
| `ExtractFunctionOutput` | `column-form.ts` |
| `CLICommandDefinition` | `wirings/cli/cli.types.ts` |
| `RequestHeaders`, `HTTPFunctionsMeta`, `HTTPWiringMiddleware` | `wirings/http/http.types.ts` |
| `JsonRpcError`, `getMCPResources`, `getMCPPrompts` | `wirings/mcp/mcp-runner.ts` |
| `TriggerSourceInfo` | `wirings/trigger/pikku-trigger-service.ts` |
| `onGraphNodeComplete`, `InputRef` | `wirings/workflow/graph/` |

The `*Meta` MCP accessors (`getMCPResourcesMeta`, `getMCPPromptsMeta`) are the ones `@pikku/modelcontextprotocol` actually uses; those stay.

### `isRef` looked like the twelfth, and isn't

`isRef` is the type guard that reads what `createRef` writes — the `__isRef` brand marking a graph node input as "substitute another node's output here". Nothing imported it because neither it nor `RefValue` was reachable from any entry point, so the one consumer that needed it — the inspector's graph serializer — had reimplemented the same four conditions privately as `isRefValue`, alongside its own structural copy of `RefValue`. Deleting `isRef` would have made that duplicate permanent, with the brand's shape asserted in two places free to drift apart.

So `isRef` and `RefValue` are exported from `@pikku/core/workflow` instead, and `packages/inspector/src/utils/workflow/graph/serialize-workflow-graph.ts` imports them rather than keeping its own copy.

### Verification

| check | result |
|---|---|
| `@pikku/core` full suite | 2000 pass, 0 fail, 10 skipped |
| `@pikku/inspector` full suite | 660 pass, 0 fail |
| `@pikku/core` + `@pikku/inspector` `tsc` | clean |
| tree-wide grep for each deleted name | no references outside the deletions themselves |

The non-deletion diff in this commit is Prettier reformatting that the tree already wanted — no test behaviour changes.

## Summary by CodeRabbit

* **New Features**
  * Added shared workflow reference utilities for consistent workflow inspection and serialization.
  * Added stronger permission typing for authentication-based permissions.

* **Breaking Changes**
  * Removed several deprecated and unused public exports, including legacy permission helpers and types.
  * Permission wiring now uses the streamlined permission APIs.

* **Bug Fixes**
  * Pinned the HTTP server version during CLI setup to prevent compatibility issues.

* **Chores**
  * Applied formatting and type-safety improvements without changing runtime behavior.

### DIFF
--- a/.changeset/delete-unreferenced-exports.md
+++ b/.changeset/delete-unreferenced-exports.md
@@ -1,0 +1,14 @@
+---
+'@pikku/core': patch
+'@pikku/inspector': patch
+---
+
+Delete exports nothing references
+
+A sweep of the `@pikku/core` surface for exports with no consumer anywhere in the repo — no package, template, verifier or e2e project imports them: `ExtractFunctionOutput`, `CLICommandDefinition`, `RequestHeaders`, `HTTPFunctionsMeta`, `HTTPWiringMiddleware`, `JsonRpcError`, `TriggerSourceInfo`, `getMCPResources`, `getMCPPrompts`, `onGraphNodeComplete` and `InputRef`.
+
+Every one was a compatibility promise with nothing on the other end of it. Removing them narrows what 0.13 has to keep stable.
+
+`isRef` looked like the twelfth, and isn't. It is the type guard that reads what `createRef` writes — the `__isRef` brand marking a graph node input as "substitute another node's output here". Nothing imported it because neither it nor `RefValue` was reachable from any entry point, so the one consumer that needed it, the inspector's graph serializer, had reimplemented the same four conditions privately as `isRefValue` along with its own structural copy of `RefValue`. Deleting `isRef` would have made that duplicate permanent, with the brand's shape asserted in two places free to drift apart.
+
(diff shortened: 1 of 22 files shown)


yasserf:
@coderabbitai review

Issue #1182, opened 2026-08-09
core: split pikku-workflow-service into composable modules


yasserf:
`pikku-workflow-service.ts` has grown past the point where the pieces can be read or tested apart: run lifecycle, step execution, sleep/resume, and store interaction all live in one class body.

Split it into composable modules with the same external behaviour, and keep it under the size ceiling.

Fifth in the core surface stack, after #1181.

PR #1188, opened 2026-08-09
refactor(core): assert to named types, never to any

### LINKED ISSUE #1180: core: assert to named types, never to any
`@pikku/core` carries 108 `as any` casts. Each one is a place the compiler stops checking, and several sit on the boundary between a wire and the runtime — exactly where a type error is most expensive to find later.

Replace every one with an assertion to a named type, or with a change to the surrounding types that makes the assertion unnecessary. Target: 0.

Third in the core surface stack, after #1179.

yasserf:
Closes #1180.

**Third in the core-surface stack.** Targets #1187 — review that one first; this diff is only the third commit.

`@pikku/core` carried **108 `as any` casts** across its non-test modules. This takes that to **zero**. Every one is replaced by an assertion to the type actually wanted, or by a change to the surrounding types that makes the assertion unnecessary.

A cast to `any` is not a neutral annotation — it discards the target type, so a wrong value passes as readily as a right one. Three of these were doing exactly that:

### `AIStreamEvent.approval-request` declared `runId` optional

`AIAgentResult['pendingApprovals']` has always required it, and every emitter already set it. The optional let `undefined` reach the field a consumer uses to resume a suspended run.

### `wireChannel` registered two handlers uncallable

`CoreChannel` accepts three handler shapes, including a wrapper around a function config (`{ func: { func }, middleware }`). `wireChannel` unwrapped it for `onMessage` and `onMessageWiring` but registered `onConnect` and `onDisconnect` as-is — the registered config's `func` was an object, and the function runner threw when it tried to call it.

Its type arguments were also misaligned with `CoreChannel`'s parameter list: `PikkuPermission` was passed into the `ChannelConnect` slot and `PikkuMiddleware` into `ChannelDisconnect`. End users never saw it — the CLI's generated wrapper casts before calling core — but anyone importing `wireChannel` from `@pikku/core/channel` directly did.

The three handler slots stay generic, now in the right positions. They have to: a wired handler's `services` parameter is narrower than `CoreSingletonServices` — its `schema` is required, not optional — and parameter types are contravariant, so pinning the slots to their `CorePikkuFunction<unknown, unknown>` defaults rejects every real handler. `verifiers/addon` calls `wireChannel` directly and is the case that proves it. `PikkuPackageState`'s channel map widens to match.

### `wire.logger` was read but never declared

The no-op audit service already preferred `wire.logger` over the singleton, so a host had no typed way to attach an invocation-scoped logger to something that would have used it.

### Holding the line

`no-any-casts.test.ts` walks the package's non-test modules and asserts the offender list is empty. It also asserts its **own matcher** works — that it finds `y as any` and `foo(bar as any)`, and ignores prose like "a picture, as any URL a browser can load". A regex that matched nothing would otherwise pass forever.

### Five casts the source history fixed elsewhere

Two in `graph-runner.ts` and three across `pikku-workflow-service.ts` / `pikku-scenario-service.ts` were cleaned up as a side effect of the workflow-service split (#1182, later in this stack). Since *this* PR is the one that promises zero, they are narrowed here instead — `(value as { $ref?: unknown })`, `(error as { code?: string })`, and dropping a cast that `PikkuRawWire = Omit<PikkuWire, 'rpc'>` already makes unnecessary.

### One call site had to catch up

`startWorkflow`'s `rpcService` parameter was `any`; it is now `PikkuRPC`. The CLI's scenario runner passes a deliberate sentinel — an RPC whose every member throws, so that a `workflow.do` without `{ actor }` fails loudly instead of quietly running against local services rather than the target environment. As a partial object it only satisfied `any`.

It is now a complete `PikkuRPC`, every member routed through one `refuseInternal` that throws the same explanatory error. That is the point of the change rather than a cost of it: the sentinel now has to state that it stands in for the whole surface, and if the surface grows a member the sentinel does not answer, the build says so.

`rpcWithWire` is named in an intersection alongside `PikkuRPC` because the workflow runner reaches for that member through an internally untyped path, and the type that declares it publicly only arrives with #1190 later in this stack.

### Verification

| check | result |
|---|---|
| `@pikku/cli` `tsc` | clean |
| `yarn build:packages` (the CI Build step) | exits 0 |
| `no-any-casts` / `channel-handler-shapes` / `gateway-channel-meta` | 10/10 |
| `verifiers/addon` (`pikku` + `tsc` + contract assertions) | 7/7 |
| grep for `as any` in non-test core source | 2 hits, both prose the matcher excludes by design |

One run of the full suite failed at `crypto-utils.test.ts:469` — the KEK derivation test that budgets 5ms of wall clock. That is the load-dependent flake **PR #1175 fixes**; it is not on this branch and is unrelated to this change. It passed on re-run.

### DIFF
--- a/.changeset/fix-channel-handler-wrapper-registration.md
+++ b/.changeset/fix-channel-handler-wrapper-registration.md
@@ -1,0 +1,9 @@
+---
+'@pikku/core': patch
+---
+
+Fix `onConnect`/`onDisconnect` wrapper handlers, and correct `wireChannel`'s generic arguments
+
+`CoreChannel` accepts three shapes for a handler: a function config, a simple wrapper (`{ func, middleware }`), and a wrapper around a function config (`{ func: { func }, middleware }`). `wireChannel` unwrapped the third shape for `onMessage` and `onMessageWiring`, but registered `onConnect` and `onDisconnect` as-is — so the registered config's `func` was an object, and the function runner threw when it tried to call it. All three shapes now register a callable config on every handler.
+
+`wireChannel`'s type arguments were also misaligned with `CoreChannel`'s parameter list: `PikkuPermission` was being passed into the `ChannelConnect` slot and `PikkuMiddleware` into `ChannelDisconnect`, which made `wireChannel({ onConnect: { func } })` fail to typecheck with "not assignable to `CorePikkuPermission`". End users did not see this — the CLI's generated `wireChannel` wrapper casts before calling core — but anyone importing `wireChannel` from `@pikku/core/channel` directly did. The signature is now `wireChannel<In, Channel>(channel: CoreChannel<In, Channel>)`; the three surplus type parameters were never used for anything correct and are gone.
--- a/.changeset/tighten-types-that-any-was-masking.md
+++ b/.changeset/tighten-types-that-any-was-masking.md
@@ -1,0 +1,11 @@
+---
+'@pikku/core': patch
+---
+
+Tighten several public types that `as any` was masking
+
+- `AIStreamEvent`'s `approval-request` variant now declares `runId: string` rather than `runId?: string`. Every emitter already set it, and `AIAgentResult['pendingApprovals']` has always required it — the optional let `undefined` reach a field consumers rely on to resume a suspended run.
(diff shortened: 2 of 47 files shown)


yasserf:
@coderabbitai review

PR #1189, opened 2026-08-09
refactor(core): give each wire an explicit set of crossovers

### LINKED ISSUE #1181: core: give each wire an explicit set of crossovers, and break the rpc→ai-agent edge
The wires (http, channel, queue, scheduler, cli, rpc, ai-agent, workflow) import each other ad hoc. Nothing declares which crossovers are intended, so an accidental import is indistinguishable from a designed one — and one of them, `rpc → ai-agent`, pulls the whole agent runtime into any program that only wanted RPC.

Two parts:

1. Give each wire an explicit, pinned set of crossovers, so an unintended edge is a test failure rather than a code review catch.
2. Break the `rpc → ai-agent` edge, and split the conformance suite that was the reason it existed.

Fourth in the core surface stack, after #1180.

yasserf:
Closes #1181.

**Fourth in the core-surface stack.** Targets #1188 — review that one first; this diff is only the fourth commit.

### Every wire declares its crossovers

The eight wires (`http`, `channel`, `queue`, `scheduler`, `cli`, `rpc`, `ai-agent`, `workflow`) imported each other ad hoc. Nothing recorded which edges were intended, so an accidental import looked exactly like a designed one.

Each wire now declares the crossovers it is allowed. `wirings-stay-decoupled.test.ts` walks the import graph and fails on:

- an edge that is not declared, and
- a **declared crossover that no longer exists** — so the declaration cannot rot into a rubber stamp that permits everything,

plus a self-check that the walker finds the edges it is meant to.

### The agent facade moves next to the agent runtime

`wire.rpc.agent` — run, stream, resume, interrupt, approve — was implemented inline in `rpc-runner.ts`, which put the agent turn logic inside the RPC primitive. It moves to `ai-agent/agent-rpc.ts`, beside the runner and stream code it delegates to, and the turn logic the runner needed in order to be importable moves into `ai-agent-turn.ts`.

`rpc-runner` imports it normally. `agent` stays a getter, so a request that never touches an agent never builds the facade — which is also why the per-invocation RPC view is a class and not an object literal: an accessor on a literal is defined per instance, dropping it off V8's fast construction path at 1.15µs against 0.47µs, on every request. That one is recorded in [the per-invocation rpc view is a class](packages/core/knowledge/decisions/internals/the-per-invocation-rpc-view-is-a-class.md).

### The conformance suite splits by service

`service-tests.ts` was 1216 lines covering ten services. It is now one module per service, so a backend author can read the contract for the one they implement. `defineServiceTests` is unchanged for callers — it dispatches to whichever the caller supplied.

### Verification

| check | result |
|---|---|
| `@pikku/core` full suite | 2027 pass, 0 fail, 10 skipped |
| `wirings-stay-decoupled` / `production-barrels-stay-lean` | 7/7 |
| `api-report.md` regenerated | no drift |

`unsupportedChannelRemote` moves from `channel-rpc-service.ts` to `channel-rpc.types.ts`, alongside the error it throws. It is still exported from `@pikku/core/channel`, so both verifier consumers (`middleware-and-permissions`) are unaffected.

### Conflicts resolved while porting

This is a patch port, not a cherry-pick, and three files needed a decision:

- `ai-agent-stream.ts` — the source commit also imported `acquireRunClaimLock` from the single-use-resume work, which is **deferred to #1159**. Dropped; only the `applyInputMiddleware` import is kept.
- `internals/index.md` — the source diff added four entries; two name notes that belong to the docs PR and do not exist yet on this branch. Only the note this PR actually adds is listed.
- `service-tests.ts` — the inline `SessionStore` block carried the `AppSession` tightening from #1188. Verified it survived into the extracted `session-store-tests.ts` rather than being lost in the split.


yasserf:
@coderabbitai review

Issue #1184, opened 2026-08-09
core: pin the public API at member level, and state what it is


yasserf:
Nothing currently catches a change to the published surface of `@pikku/core`. An export can be added, removed, or have a member's signature changed, and the only signal is a downstream break after release.

Three parts:

1. Pin the export surface against drift.
2. Pin it at **member** level, not just export names — a changed method signature is a change to the API.
3. Make the generated report state what the API *is*, not merely list identifiers, so a reviewer can read the diff and understand what changed.

Also declare which modules have import-time side effects, so bundlers can drop the rest.

Last in the core surface stack (after #1183) because it owns the generated artifacts — `public-surface.json`, `api-report.md` and the core `package.json` exports map — which every earlier PR in the stack leaves for this one to regenerate once.

PR #1190, opened 2026-08-09
refactor(core): split pikku-workflow-service into composable modules

### LINKED ISSUE #1182: core: split pikku-workflow-service into composable modules

yasserf:
Closes #1182.

**Fifth in the core-surface stack.** Targets #1189 — review that one first; this diff is only the fifth commit.

`pikku-workflow-service.ts` had reached **2325 lines**, with the run lifecycle, error catalog, run-engine interfaces, queue routing, queue wiring, approvals and recovery all in one module. It is now **1993**, and the pieces live where they can be read and tested apart:

| module | what moved |
|---|---|
| `workflow-errors.ts` | the error catalog |
| `workflow-run-engine.types.ts` | run-engine interfaces |
| `workflow-constants.ts` | timing and limit constants |
| `workflow-meta-resolver.ts` | meta lookup |
| `workflow-queue-routing.ts` | queue and group name resolution |
| `workflow-queue-wiring.ts` | worker wiring |
| `workflow-approval.ts` | the approval path |
| `workflow-recovery.ts` | stalled-run and undispatched-step recovery |

`@pikku/core/workflow` exports the same names, and importers point at the real modules rather than re-export shims. `source-files-stay-composable.test.ts` holds a **2000-line ceiling for every module in the package** — the split is only worth doing if it stays done.

## Two defects the split surfaced

Both were held in place by an `any`. Typing the meta resolver and the `rpcService` parameter is what exposed them, which is why they are fixed here rather than separately.

### A run with no generated metadata crashed instead of saying so

It threw `TypeError: Cannot read properties of undefined` from deep inside the runner, or `WorkflowNotFoundError` — neither of which points at the cause. It now throws `PikkuMissingMetaError`, matching how the queue and trigger runners already report the same condition.

### A child workflow ran as nobody

The service filled its wires from `rpcService.wire?.session`, `.rpc` and `.pikkuUserId`. **`PikkuRPC` has no `wire`**, and neither does the object the RPC service actually returns — so every one of those reads was `undefined`, and `rpcService: any` is what kept it quiet. The visible consequence: a child workflow started from a step ran with no user.

Those reads now come from the run record, which is durable and survives the process boundary a queued step crosses. `session` and `rpc` are not copied at all — `runPikkuFunc` attaches `rpc` lazily per invocation and resolves the session from the store, so both were overwritten moments later regardless.

`PikkuRPC` also now declares `rpcWithWire`, which the RPC service has always returned and the workflow service has always called.

[A knowledge note](packages/core/knowledge/decisions/internals/a-workflow-wire-is-built-from-the-run-not-from-the-rpc-service.md) records the reasoning, and is added to the internals index in filename order.

### Verification

| check | result |
|---|---|
| `@pikku/core` full suite | 2024 pass, 0 fail |
| `source-files-stay-composable` / `workflow-missing-meta` / `workflow-child-run-session` / `no-any-casts` | 9/9 |
| largest module in the package | 1993 lines, under the 2000 ceiling |

`public-surface.json` is untouched — the API pin PR (#1184) owns the generated artifacts and regenerates them once, at the end of the stack.

### DIFF
--- a/.changeset/child-workflows-inherit-the-parent-user.md
+++ b/.changeset/child-workflows-inherit-the-parent-user.md
@@ -1,0 +1,11 @@
+---
+'@pikku/core': patch
+---
+
+A child workflow now inherits the `pikkuUserId` its parent was running as
+
+`PikkuWorkflowService` filled parts of the wire it builds for a workflow body and for a child run from the RPC service it had been handed — `rpcService.wire?.session`, `rpcService.wire?.rpc`, `rpcService.wire?.pikkuUserId`. `PikkuRPC` has no `wire`, and neither does the object the RPC service actually returns, so every one of those reads was `undefined`; the `rpcService: any` parameter type is what kept it quiet. The visible consequence was that a child workflow started from a step ran as nobody.
+
+Those reads now come from the run record, which is durable and survives the process boundary a queued step crosses. `session` and `rpc` are not copied at all: `runPikkuFunc` attaches `rpc` lazily per invocation and resolves the session from the session store, so both were overwritten moments later regardless. The wires these paths build are typed `PikkuRawWire`, which is what they have always been.
+
+`PikkuRPC` also now declares `rpcWithWire`, which the RPC service has always returned and the workflow service has always called.
--- a/.changeset/tidy-workflow-service-split.md
+++ b/.changeset/tidy-workflow-service-split.md
@@ -1,0 +1,9 @@
+---
+'@pikku/core': patch
+---
+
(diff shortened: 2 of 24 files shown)


yasserf:
@coderabbitai review

PR #1191, opened 2026-08-09
refactor(core)!: move the adapter surface to /ecosystem so the root can promise stability

### LINKED ISSUE #1183: core: move the adapter surface to /ecosystem so the root can promise stability
The types a runtime adapter implements against sit on the root entry point of `@pikku/core`, mixed in with the API application code uses. That makes the root impossible to stabilise: an adapter-internal change reads as a breaking change to everyone.

Move the adapter surface onto a `@pikku/core/ecosystem` entry point, leaving the root as the surface we can commit to for 0.13.

Breaking for adapter authors, appropriate pre-0.13.

Sixth in the core surface stack, after #1182.

yasserf:
Closes #1183.

**Sixth in the core-surface stack.** Targets #1190 — review that one first; this diff is only the sixth commit.

The types and helpers a runtime adapter, a service package, the code generator and the CLI implement against sat on the **package root**, mixed in with the API application code uses. That made the root impossible to commit to for 0.13 — an adapter-internal reshape reads as a breaking change to everyone.

The evidence for the split is in the history: `runPikkuFunc` was reshaped repeatedly over the last year while **every field on `PikkuWire` survived untouched**. Promising stability on both surfaces means promising the weaker of the two.

They now live on `@pikku/core/ecosystem`. The root is what 0.13's compatibility promise covers.

### Why that name

**Not `/internal`.** Generated bootstrap files import from here, so the specifier lands in the user's own `.pikku` directory. Telling someone they are touching internals when the code generator put it there is both wrong and self-defeating — it could never be broken anyway.

**Not `/runtime`.** That reads as runtime-versus-compile-time, i.e. the real API; `packages/runtimes/*` already claims the word; and the CLI is the largest consumer of this surface, not a runtime.

`./internal` stays as an **alias to the same module**, because the pinned bootstrap CLI still emits it.

The reasoning is recorded in [a knowledge note](packages/core/knowledge/decisions/internals/the-ecosystem-entry-point-carries-the-adapter-surface.md).

Breaking for adapter authors; appropriate pre-0.13.

### Scope note

In the source history this surface moved three times — to `/internal`, then `/runtime`, then `/ecosystem` — as the naming got argued out. Only the destination is useful, so this PR is the **net effect**: one move, to the name that was landed on. The two intermediate changesets are replaced by one describing where it actually ended up.

### Verification

19 packages import the moved surface: the CLI, the inspector, both deploy adapters, nine runtimes, four services, `schedule` and the console addon.

| check | result |
|---|---|
| `@pikku/core` full suite | 2024 pass, 0 fail |
| `@pikku/core` `tsc` + build | clean |
| `tsc` across all non-template/verifier workspaces | **zero** errors mentioning `@pikku/core` |
| `@pikku/node-http-server` | 26/26 |
| `@pikku/express` | 3/3 |
| `@pikku/bun-server` | 5/5 (17 skipped — needs Bun) |
| `@pikku/schedule` | 1/1 |

The remaining repo-wide `tsc` errors are the **pre-existing** codegen blocker (`Cannot find module '#pikku…'` / `'../.pikku/…'` and the implicit-`any` cascade behind it), reproduced on a pristine `main` checkout. None of them names a `@pikku/core` specifier.

### Conflicts resolved while porting

Four files overlap with earlier PRs in this series that are **not** in this stack:

- `in-memory-scheduler-service.ts` / `.test.ts` — the source also carried the delayed-RPC fix from #1169. Only the import path is changed here.
- `pikku-express-server.test.ts` — the source also carried the upload-signature test from #1177. Only the import path is changed here.
- `bench-profile-granular.ts` / `bench-profile-hotpath.ts` — the source also carried the relative-path repair from #1173. Only the import path is changed here.

`internals/index.md` also wanted two entries that belong elsewhere: one names a note that does not exist on this branch, the other duplicates an entry already present. Neither was taken.

`public-surface.json` is untouched — the API pin PR (#1184) owns the generated artifacts.


yasserf:
@coderabbitai review

PR #1192, opened 2026-08-09
feat(core): pin the public API at member level, and state what it is

### LINKED ISSUE #1184: core: pin the public API at member level, and state what it is

yasserf:
Closes #1184.

**Last in the core-surface stack.** Targets #1191 — review that one first; this diff is only the seventh commit.

Nothing caught a change to what `@pikku/core` publishes. An export could appear, vanish, or have a member's signature change, and the only signal was a downstream break *after* release.

### Two pins, at two levels

**`public-surface.json`** pins every runtime export reachable through a `package.json` `exports` subpath — 49 entry points. It catches an export appearing or vanishing, including the ones an `export *` widens by accident.

**`api-report.md`** pins the API at **member** level. A method added to an interface is the change that actually breaks a consumer's build, and an export list cannot see it. The report is written to state what the API *is* rather than merely list identifiers, so the diff is readable by a reviewer rather than a wall of names.

Both regenerate from the code and are asserted against it. `api-report.test.ts` regenerates and requires no diff, and separately asserts the report **contains member-level detail** — a generator that quietly degraded to names would otherwise keep passing.

Types are not in `public-surface.json`: they are erased before the test can enumerate them. That is what `api-report.md` is for.

### `sideEffects` is an allowlist, not `false`

Core genuinely has import-time side effects: the error registry is built by `addError` calls that run on import. Claiming `sideEffects: false` would let a bundler drop those modules and leave `getErrorResponse` unable to find any error.

`side-effects-are-declared.test.ts` detects which modules actually have side effects and fails if the allowlist disagrees — in either direction — plus a self-check that the detector finds a known one.

### Why this PR is last

Every earlier PR in the stack changes the surface: #1186 removes three exports, #1187 deletes eleven more, #1189 moves `unsupportedChannelRemote`, #1190 splits the workflow service, #1191 moves the whole adapter surface to `/ecosystem`. Each of those deliberately left `public-surface.json` alone.

Both artifacts are **generated here**, once, against the final surface — rather than carried through seven conflicting versions of a file that is output, not source. Verified: `/ecosystem` is present, and every export removed earlier in the stack (`ZodLike`, `addTagPermission`, `ExtractFunctionOutput`, `HTTPWiringMiddleware`, `getMCPResources`, …) is absent from both.

### Verification

| check | result |
|---|---|
| `public-surface` / `side-effects-are-declared` | 5/5 |
| `api-report` | 2/2 — regenerating produces no diff |
| `yarn api-report` | written for 49 entry points |

### Index note

The source commit regenerated `internals/index.md` wholesale (145 lines reordered). Rather than carry a reordering that conflicts with every other PR in the stack, the current index is kept and the two new notes are inserted in filename order.


yasserf:
@coderabbitai review

PR #1193, opened 2026-08-09
docs(core): move the reasoning out of the comments and into knowledge

### LINKED ISSUE #1185: docs(core): move load-bearing reasoning into the knowledge bundle
The reliability, security and surface-tightening work across #1160–#1168 and #1178–#1184 produced a set of decisions whose *reasoning* currently lives only in commit messages and PR descriptions — the least durable place for it.

Move the load-bearing reasoning into `packages/core/knowledge/`, including why the middleware resolution cache is deliberately unbounded, and land the combined changelog and findings report for the sweep.

Final PR in the stack, after #1184.

yasserf:
Closes #1185.

**Final PR in the stack.** Targets #1192 — review the code PRs first; this one is docs and comments.

### The reasoning moves out of the comments

A twelve-line comment explaining *why* a line is the way it is has two problems: it goes stale as soon as the code around it moves, and nobody who has not already opened that file will ever read it. Each of those comments is replaced by a one-line pointer:

```ts
// knowledge: decisions/internals/only-exposed-functions-enter-a-virtual-user-catalogue.md
if (meta.expose !== true) continue
```

**34 new notes** across `decisions/internals`, `decisions/security` and `questions`, covering the agent runtime, virtual users, actor flow, voice input/output, and the middleware resolution cache.

Two are **questions**, not decisions — behaviour that looks wrong and has not been settled:

- `channel-middleware-accepts-bare-factories-that-nothing-resolves.md`
- `unauthorized-channel-replies-escape-the-declared-out-type.md`

### The three summary documents

`CHANGELOG.md`, `CHANGELOG.verbose.md` and `RELIABILITY-FINDINGS.md` — the write-ups from the review this whole stack came out of. `RELIABILITY-FINDINGS.md` in particular records findings that were *not* fixed here, so they are not lost when the branches merge.

### No changeset

Nothing in this PR changes behaviour or the API. The only non-comment source edit is a JSDoc line that described inline step metadata as "legacy support" when nothing has replaced it. Cutting a `@pikku/core` patch release for comment reflow would be noise.

### Index note

`internals/index.md` conflicted in every PR in this stack that touched it. Rather than carry the source's wholesale regeneration, the entries from both sides were unioned, deduped by target filename and sorted: 132 → 115. Verified lossless — nothing that HEAD listed is absent, 27 entries added, and every listed target now resolves to a file that exists.

(24 note files in that directory remain unlisted in the index. They are unlisted on `main` too — pre-existing, and out of scope here.)

### Verification

| check | result |
|---|---|
| `prettier --check` on changed source | clean |
| index integrity | 115 listed, 0 dangling |

Knowledge `.md` files are not prettier-formatted anywhere in this repo — files this PR does not touch warn identically on `main`.


yasserf:
@coderabbitai review

PR #1400, opened 2026-08-22
fix(better-auth): gate actor sign-in on the command, not on a secret being set

yasserf:
`actor()` decided whether to work by asking whether a secret was configured, which is wrong in both directions.

In development it charged every contributor a hand-managed `SCENARIO_ACTOR_SECRET` for a server that is never production, and skipping it rendered no quick-login control for no stated reason. In production it did the opposite: a secret that leaked into a deployed environment silently enabled passwordless sign-in as any declared persona, including the admin ones.

The gate now reads `PIKKU_DEV_ACTOR_SIGN_IN`, a positive marker `pikku dev` sets and `pikku serve` clears — deliberately not `NODE_ENV`, which is written by bundlers, test runners and process managers and is simply absent in plenty of real deployments, so a `!== 'production'` test fails *open* exactly where being wrong costs something. A marker fails closed.

`pikku dev` mints an ephemeral per-run secret when none is set, under both `SCENARIO_ACTOR_SECRET` and the `VITE_`-prefixed copy the dev frontend reads; an explicitly-set secret still wins.

## One hatch, and it is an environment variable

Scenario suites against a deployed stage opt in with `PIKKU_ALLOW_ACTOR_SIGN_IN=passwordless-actor-sign-in` — a sentence rather than a boolean, so it cannot be reached by habit, and any other value is ignored *and* warned about.

There is deliberately no build-time equivalent. An `allowOutsideDev` option would be compiled into the bundle, so nothing about a running deployment would tell you whether its actor endpoint is open, and `pikku serve` could clear a marker but never a constant. An env var is the auditable form: greppable across deployments, visible where the deployment is configured, clearable.

## Signing in and provisioning are separate powers

Creating an `actor: true` row for an address that has none requires `pikku dev` specifically. The opt-in permits authentication only, refusing an unknown address with `No actor account exists for that address`. So a stage that must run scenarios can be signed into as the personas provisioned there without also becoming a place where anyone holding the secret can mint identities.

The endpoint was the only thing that created actor rows, so `pikku persona sync` now writes them through the database connection it already holds and already reads `user` from. Three consequences:

- **It needs no `SCENARIO_ACTOR_SECRET`** — provisioning never touches the API, so a stage can be provisioned before anyone decides whether its actor endpoint should open at all.
- **A real user's address is refused** rather than being granted the persona's roles.
- **App-level `databaseHooks.user.create` do not fire** for persona accounts. Nothing in pikku defines one; an app that does should know these rows arrive by provisioning rather than by signup.

Every branch reports which one it took, and a secret wired to a shut gate warns rather than doing nothing quietly.

## Upgrading

A deployed stage that runs scenarios needs `PIKKU_ALLOW_ACTOR_SIGN_IN` set, and its personas provisioned with `pikku persona sync <environment>`, or sign-in returns 401. Local `pikku dev` and `pikku scenario run --spawn` need nothing.

## Tests

- `packages/services/better-auth`: 146/146 pass
- `packages/cli`: 1265/1265 pass

## Summary by CodeRabbit

- **New Features**
  - Added configurable actor sign-in controls for development and deployed environments.
  - Development mode enables quick actor login and synchronizes generated credentials.
  - Added deployment-time persona provisioning with account creation and additive role assignment.
  - Added orphan reporting, optional banning, session revocation, and ban restoration when personas return.

- **Changes**
  - `persona sync` now reports eligible personas and skips without modifying the database.
  - Unknown actors are blocked unless explicitly permitted in deployed environments.
  - Added clear warnings for invalid settings, credential mismatches, and refused sign-ins.

- **Documentation**
  - Updated actor authentication and provisioning guidance.


packages/cli/src/functions/commands/persona-sync.ts
```diff
@@ -19,16 +20,58 @@ import { loadDeclaredRoles, openScopeServiceForRoles } from './roles-shared.js'
...
+  return row ? { id: String(row.id), actor: Boolean(row.actor) } : null
+}
+
+/**
+ * Better Auth's own id shape: 32 alphanumeric characters. A uuid would be four
+ * characters longer than a `varchar(32)` column allows.
+ */
+const generateUserId = (): string => {
+  const alphabet = 'abcdefghijklmnopqrstuvwxyz0123456789'
+  const bytes = randomBytes(32)
```
yasserf: dont we use uuiid?

packages/cli/src/functions/commands/persona-sync.ts
```diff
@@ -19,16 +20,58 @@ import { loadDeclaredRoles, openScopeServiceForRoles } from './roles-shared.js'
...
+ * one where anyone holding the secret may invent identities.
+ */
+const provisionActor = async (
+  db: Kysely<any>,
+  email: string,
+  name: string
+): Promise<string> => {
+  const id = generateUserId()
+  const now = new Date()
+  await db
```
yasserf: er.. cli doesnt have access to the prod/staging db,

yasserf: You're right, and it applies wider than these lines: `openScopeService` resolves its connection from the local project config, so `pikku persona sync staging` read staging's API and would have written whatever database the checkout happened to point at.

Fixed in d1a1a41b7 by moving provisioning to where the database already is. `@pikku/better-auth` now exports `provisionPersonas`, called from the app's own server lifecycle:

```ts
await provisionPersonas(singletonServices, {
  personas: personaConfigs,
  environments: personaEnvironments,
})
```

It creates missing accounts through better-auth's internal adapter (so the id shape and column set stay better-auth's business), applies declared roles, and is additive. The generated personas file now emits `personaEnvironments` — just the `production` flag, not apiUrl/paths — so the same rule that decides who may *run* in an environment decides who gets provisioned there.

`pikku persona sync <environment>` stays, but reports instead of writing: who that environment will provision, with which roles, and why anyone was skipped. No database, no `SCENARIO_ACTOR_SECRET`.

Two calls worth flagging: I didn't generate a `syncPersonas(services)` wrapper into `pikku-personas.gen.ts` — every generated project imports that file and the wrapper would drag `@pikku/better-auth` into all of them — and I kept `persona sync` rather than deleting it, since it's already released.

PR #1453, opened 2026-08-25
docs(skills): machine auth is middleware that sets a session

yasserf:
A caller that has an identity — a sandbox, a deployed container, a machine host, a device, a cron — is authenticated **once**, in middleware that calls `setSession`. The function is then an ordinary `pikkuFunc` gated with `scopes`, and it reads `session`. It never re-derives who the caller is in its own body, and never in a `permissions` check that verifies a token and returns `true`.

The middleware skill also records why this cannot be `addTagMiddleware` over `/rpc`: tag middleware runs inside `runPikkuFunc`, which the RPC dispatch calls with no `sessionService`, so `invocationWire.session` is never re-read and the session is invisible to the function.

**Crons.** A scheduled task has no caller, but it is still a machine principal, and without a session it cannot invoke a gated RPC or be attributed. Its own `middleware` already does the job — `runScheduledTask` builds its wire with a `sessionService`, so a `setSession` there is what the function is frozen with. This branch first added a `session` field to `wireScheduler`; that is reverted. It was a second way to do what middleware already does, and it invited a per-cron session literal restating scopes and an org that a cross-org sweep has to ignore. A cron is a scope, not a seeded user row. One test pins the middleware path.

Net change: three skill docs, one core test, one patch changeset for `@pikku/skills`.


yasserf (commented):
Verified the load-bearing claim rather than taking it on trust. `wiringAuth` is read in exactly three places in `packages/core/src/function/function-runner.ts`: the sessionless branch tests `wiringAuth === true` (unchanged when `false` becomes `undefined`), the session-taking branch's `if (!session) throw MissingSessionError` is unconditional, and only the warn at line 336 was gated on `=== false`. So dropping `auth: false` from `runScheduledTask` really does change nothing but the spurious warning — "Nothing else changes" is accurate. Both new tests are genuine TDD; the second was red before the change.

One nit: `a session-taking task does not warn that auth was disabled` asserts *zero* warn logs rather than the absence of that specific warning, so any unrelated future warn on that path fails it for the wrong reason. Matching the message would keep the failure pointing at the right thing.

Otherwise LGTM.

PR #1457, opened 2026-08-25
fix(core): queue a workflow step that names another workflow

yasserf:
## The bug

`dispatchStep` decides whether a step goes to the queue by reading `workflowQueued` off `rpc` meta:

```ts
const rpcFuncId = pikkuState(null, 'rpc', 'meta')[rpcName]
```

`addWorkflow` registers a workflow into `workflows` meta and the function registry, never into `rpc` meta. So for a step naming another workflow the lookup always misses, `forceQueue` is always false, and a child workflow can never be queued.

It therefore always took the inline path in `rpcStep`: `startWorkflow(..., { inline: true })` followed by `awaitRunEnd`, an unbounded `while (true)` poll with no timeout — all of it inside the parent's `withRunLock`. Two consequences:

- the parent holds its run lock, and that lock's database connection, until the child ends
- the child is registered inline, so its own `workflow.sleep` stops being a durable suspension and becomes a real in-process `setTimeout`

A parent whose child polls for fifteen minutes holds two lock connections for fifteen minutes. Enough concurrent runs and the lock pool is exhausted; every other run then blocks on `pg_advisory_lock`, times out, is redelivered, and the engine stops moving while every part of it still looks healthy.

Observed on a deployment with a 16-connection lock pool: 16 advisory locks held, every session idle on `SELECT pg_advisory_lock($1)`, the shared orchestrator consumer pinned at `num_ack_pending == max_ack_pending` with a backlog climbing past 450, and 39 runs stuck — 29 parents blocked on 24 children marked `inline = true`.

## The fix

A step naming a workflow now queues whenever a queue service exists, even unmarked. That reaches `executeWorkflowStepInner`, which already starts the child with `inline: false` and throws `ChildWorkflowStartedException` — the parent acks and unwinds, and `onChildWorkflowCompleted` resumes it when the child finishes. The non-blocking path was already there; it was simply unreachable.

An inline parent still runs its children inline, and a deployment with no queue service is unchanged.

The decision moves into `stepDispatchTarget` in `workflow-queue-routing.ts`, beside the other routing helpers — `pikku-workflow-service.ts` was one line under the repo's 2000-line module cap.

## Tests

Three new cases in `pikku-workflow-service.test.ts`: a workflow step queues while unmarked; stays inline when the parent run is inline; stays inline with no queue service. Existing `workflowQueued` cases unchanged.

`packages/core`: 2322 pass, 0 fail.

https://claude.ai/code/session_01BX8tyDtWHyDfij52MjDr7f

### DIFF
--- a/.changeset/queue-child-workflow-steps.md
+++ b/.changeset/queue-child-workflow-steps.md
@@ -1,0 +1,9 @@
+---
+'@pikku/core': patch
+---
+
+Queue a workflow step that names another workflow, instead of running it inside its parent.
+
+`dispatchStep` decided by reading `workflowQueued` off `rpc` meta, but `addWorkflow` never registers there, so a child workflow could never be queued. It always took the inline path: the parent started the child with `inline: true` and then sat in an unbounded `awaitRunEnd` poll, holding its run lock and that lock's connection until the child ended — and the child, being inline, ran its own `sleep` as a real in-process wait rather than a suspension. A parent whose child polled for fifteen minutes held two lock connections for fifteen minutes, and enough of them exhausted the lock pool and stalled every other run behind it.
+
+A step naming a workflow now queues whenever a queue service exists, reaching the `ChildWorkflowStartedException` path that already unwinds the parent and resumes it when the child completes. An inline parent still runs its children inline.
--- a/packages/core/src/wirings/workflow/pikku-workflow-service.test.ts
+++ b/packages/core/src/wirings/workflow/pikku-workflow-service.test.ts
@@ -192,8 +192,11 @@ describe('pikku-workflow-service run-level inline', () => {
 
 describe('pikku-workflow-service per-function step dispatch', () => {
   class TestWorkflowService extends InMemoryWorkflowService {
-    public callDispatchStep(rpcName: string) {
-      return this.dispatchStep('run-1', 'step-1', rpcName, {})
+    public callDispatchStep(rpcName: string, runId = 'run-1') {
+      return this.dispatchStep(runId, 'step-1', rpcName, {})
+    }
+    public markInline(runId: string) {
+      this.registerInlineRun(runId)
     }
   }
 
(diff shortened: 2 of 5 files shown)


yasserf (commented):
Traced the diagnosis rather than taking the changeset's word for it, and it holds up. The inline-and-poll path is at `pikku-workflow-service.ts:1659` — reached only when `dispatchStep` returns false — and it starts the child with a hardcoded `inline: true` followed by `awaitRunEnd`. Because `workflowQueued` is read off `rpc` meta and `addWorkflow` never registers there, that branch was unreachable-by-any-other-route for every child workflow, exactly as described. With the fix, `dispatched === true` raises `WorkflowAsyncException`, a step worker picks the step up, and `executeWorkflowStepInner:1387` takes the good path that throws `ChildWorkflowStartedException`. Also checked the queue name: `dedicatedQueueName` falls back to the shared `pikku-workflow-step-worker`, so a workflow-named step is not routed to a queue with no worker on it.

Two things worth a look:

**`isInline` is now awaited on every step dispatch.** `stepDispatchTarget(rpcName, stepName, await this.isInline(runId))` evaluates the parent's inline state before the function decides whether it matters. For the common case — a plain function step, not `workflowQueued`, not a workflow — the old code returned after two synchronous map lookups; now every dispatch can reach `getRunIdentity`, a store read, whenever `runContexts` has no cached entry for the run. Passing a thunk, or testing `isWorkflow` before resolving it, keeps the hot path synchronous.

**The tests cover the routing decision but not its consequence.** The three new cases assert the `dispatched` and `queued` booleans. What actually changed for every existing app with a `queueService` is that a child workflow now unwinds its parent and resumes later instead of returning its output inline — and nothing here proves the parent comes back with the child's output. The changeset says that path "already" works; a test that runs a parent with a queued child end-to-end and asserts the step result is the one that would catch it regressing.

Neither blocks merge as far as I'm concerned. The changeset is patch and `@pikku/core` is a real workspace package.

PR #1458, opened 2026-08-25
feat: separate a project's three languages — identifiers, meta, and product UI

yasserf:
## Three languages, not one

A Pikku project has three separate human-language decisions, and until now
nothing in the repo said so. An agent that met one of them assumed it had met
all three.

| Axis | What it covers | What sets it |
| --- | --- | --- |
| **Identifiers** | Function, component, type, variable and file names. Database tables and columns. Commit messages. | **Nothing — always English.** There is no setting, and this PR does not add one. |
| **Meta** | The prose authored *inside* the code: `description` on functions and steps, `name`/`title` on features and scenarios, step `template`, role and persona descriptions. | **New:** `locale` in `pikku.config.json`, default `en`. |
| **Product UI** | Every string the app shows a user. | `messages/<locale>.json`, with `active.json`'s `defaultLocale` choosing what a first-time visitor opens in. `baseLocale` stays `en`. |

## The failure this comes from

A real build, via the fabric template. The brief for a German medical practice
said *"the entire UI is German, no English strings visible anywhere"*. The agent
read one sentence about the product's users as an instruction about the
codebase and produced:

- `project.inlang/settings.json` with `baseLocale: "de"`, `locales: ["de"]`, and
  no `en.json` at all — so `fabric i18n --add-locale` has no catalogue to
  translate from, permanently. The app can never gain a second language without
  re-authoring every message key.
- RPC functions `getUebersicht`, `getPatientendetail`
- React components `Zeitstrahl`, `AufmerksamkeitStreifen`
- helpers `datumDeutsch`, `voraussichtlichFertig`
- scenario steps `arbeitslisteZeigtPatient`, `filterZeigtNurStatus`
- database tables `vorgang`, `ereignis` with German columns

The German UI it was asked for needed none of that. It needed German *values* in
a catalogue whose keys and source stayed English. What that project actually
wanted was three settings, one per axis:

```jsonc
// project.inlang/settings.json — the message source stays English
{ "baseLocale": "en", "locales": ["en", "de"] }
// apps/app/src/i18n/active.json — what a visitor opens in
{ "defaultLocale": "de" }
// pikku.config.json — the language the team reads their Console in
{ "locale": "de" }
```

It is partly the framework's fault for never asking, and the template's own
`i18n/config.ts` already documented the correct model in a comment that nothing
else in the toolchain repeated.

## What the config field does

`locale` on `PikkuCLIInput`, defaulting to `en`, validated through
`Intl.getCanonicalLocales` — Node's own BCP-47 grammar, which catches the
mistake people actually make (`de_DE` with a POSIX underscore) and returns the
canonical spelling so `EN-gb` and `en-GB` are one value downstream rather than
two that never compare equal.

**It exists for the Console.** Meta is the one part of a project the Pikku
Console renders back to a human — `ScenarioSection`, `FeatureDocument`,
`LadderStep`, `PersonaDetail` and `FunctionsListPanel` all pass it through
`asI18n()` untouched while translating their own chrome — so a team whose
working language is German should be able to read their own Console in German
without anything else about the project changing. That rationale is in the doc
comment, because a field whose purpose is unstated gets read as whatever the
reader already had in mind.

## What the skills now enforce

- **`pikku-concepts`** — new *What Language You Write In*: the canonical
  three-axis statement, the identifiers rule, and the German-portal failure told
  in full. It loads first in every Pikku task, which is why it holds the primary
  copy.
- **`pikku-build-app`** — a language question in §1, and a new **§1a** at the
  requirements-gathering point that settles all three axes and writes `locale`
  into `pikku.config.json` *before there is any meta to be wrong*. Default `en`
  without asking when the project is obviously English. The Agent Operating
  Procedure now says to read the field on every subsequent run.
- **`pikku-scenario`** — new *What language the prose is in*: the step's `name`
  is a `pikkuFuncId` the generated step map is keyed by and is English; its
  `description`/`template` follow `locale`. Plus a red-flag row.
- **`pikku-i18n`** — new *The product's language is not the code's language*:
  why `baseLocale` stays `en`, what `defaultLocale` is for, the German portal
  worked correctly, and three new *What NOT to do* bullets. A project already in
  the broken state is called out as a re-key someone has to agree to, not a fix
  to slip in.
- **`pikku-build-quick`**, **`pikku-build-platform`**, **`pikku-feature`**,
  **`pikku-fabric`** — the short form, at each one's own naming or
  requirements moment.

## The mechanical check

`pikku fabric validate` already walked every `apps/*/project.inlang/settings.json`
to read `baseLocale` (it needs the base catalogue to spot scenario steps that
hardcode copy). It read the value and never questioned it. New finding
**`app-base-locale-not-english-<app>`** (warn) does, naming `defaultLocale` as
the setting that was actually wanted and showing both edits. No new CLI surface.

## Design caveat, stated plainly

**A field called `locale` risks being read as the app's UI language** — which is
precisely the confusion that caused the original bug, so it deserves saying out
loud rather than burying. `metaLocale` would read less ambiguously against the
i18n concepts already in the template (`baseLocale`, `defaultLocale`,
`supportedLocales`), and I would mildly prefer it. I kept `locale` because that
is what was asked for, and mitigated three ways: the doc comment spends more of
its length on what the field is *not* than on what it is; the validation error
does the same, so even the failure path teaches the distinction; and every skill
that mentions it states the three axes rather than the one. Happy to rename in
review — it is a one-line change plus the prose.

## Two things in the framing that turned out not to match this repo

- **There is no `prompt.md` here.** Nothing in this tree is served at
  `pikkufabric.com/prompt.md`; `find . -name "prompt.md"` returns nothing, and
  the fabric repo has no such route either. The build instructions an agent
  actually loads are the skills in `packages/skills/skills/`, so that is where
  all of this went.
- **The fabric app template is not in this repo.** `--add-locale`,
  `--default-locale`, `apps/app/src/i18n/active.json` and the `i18n/config.ts`
  comment all live in a `starter-template` repo cloned at runtime by
  `packages/create`. The only `active.json` in this tree is the *theme*
  selector (`packages/mantine-theme/active.json`, shaped `{ "id": … }`) — worth
  knowing, since the two are easy to confuse. The skills here describe the
  template's mechanism accurately; no template file is changed by this PR.

## A limitation this PR does not fix

`composeStepProse` frames every scenario sentence in English —
`` const subject = actor ? `the ${actor}` : '' `` plus a capitalised Given/When/
Then. The Console translates the keywords into its own UI language; the CLI
reporter does not, and nothing translates `the`. So `locale: "de"` today gives
German step prose inside an English frame: `Given the shopper kauft 1 Äpfel`.
The `pikku-scenario` skill says this outright rather than letting an author
discover it. Related: where meta is absent the Console falls back to
`toEnglishName()`, which camelCase-splits the *identifier* into an
English-looking label — so under a non-`en` locale, meta is worth authoring
rather than left to the fallback.

## Verification

Run from the worktree. `packages/cli` tests use the package's own runner.

```
$ node --no-memory-protection-keys --import tsx --test src/utils/pikku-cli-config.test.ts
  ✔ locale defaults to English when the config does not name one
  ✔ a declared locale survives the load, canonicalized
  ✔ a locale that is not a language tag names itself in the error
  ✔ an absent locale is English rather than an error
  ✔ canonicalizes so one language is one value downstream
  ✔ rejects the POSIX underscore people reach for
  ✔ rejects what is not a tag at all
  ✔ the error points at the two settings it is not
ℹ tests 20
ℹ pass 20
ℹ fail 0
```

8 new tests (3 through the loader, 5 on `normalizeLocale`), covering the
default, canonicalization, and the validation error.

Type check, diffed against an untouched `origin/main` worktree sharing the same
`node_modules`:

```
$ tsc --noEmit -p packages/cli/tsconfig.json   # both trees, output sorted and diffed
identical to baseline (284 pre-existing errors, 0 new)
```

Full `packages/cli` suite, same comparison: **962 pass / 34 fail** on this
branch against **954 pass / 34 fail** on `origin/main` — the 8 added tests all
pass and the failure count is unchanged.

**Honest gap:** the 34 pre-existing failures, and the three new
`app-base-locale-not-english` tests in `validate.function.test.ts`, come from
the same environment limitation — a fresh worktree has no generated+built
`packages/cli/dist/.pikku/*/index.js`, so `#pikku/function` cannot resolve and
the file aborts before any test runs. It fails identically on untouched
`origin/main`. **Those three tests are written but I could not execute them**,
and I am not claiming them green. `tsc` does parse and type-check the body of
`validate.function.ts` (its only error is the unresolvable generated import, in
both trees), so the new block type-checks — but it has not been run.

### DIFF
--- a/.changeset/meta-has-a-language-identifiers-do-not.md
+++ b/.changeset/meta-has-a-language-identifiers-do-not.md
@@ -1,5 +1,5 @@
 ---
-'@pikku/cli': minor
+'@pikku/cli': patch
 ---
 
 `pikku.config.json` takes a `locale`, naming the language the project's meta is
@@ -9,9 +9,14 @@ Meta is the human-readable prose authored inside the code rather than in a
 message catalogue: `description` on functions and steps, `name`/`title` on
 features and scenarios, step `template` strings, role and persona descriptions.
 It is also the one part of a project the Pikku Console renders back to a human,
-which is the entire reason the field exists — a team whose working language is
-German should be able to read their own Console in German without anything else
-about the project changing.
+which is what the field is for — a team whose working language is German should
+be able to read their own Console in German without anything else about the
+project changing.
+
+This release is the groundwork: the field is read, validated and canonicalized,
+and the skills explain which of a project's three languages it is. Nothing
+consumes it yet, so setting it does not change what the Console renders — that
+comes with the reader.
 
 It defaults to `en`, is validated as a BCP-47 tag through
 `Intl.getCanonicalLocales` (so `de_DE` fails at the line that is wrong rather
--- a/.changeset/meta-has-a-language-identifiers-do-not.md
+++ b/.changeset/meta-has-a-language-identifiers-do-not.md
@@ -1,0 +1,26 @@
+---
+'@pikku/cli': minor
+---
+
+`pikku.config.json` takes a `locale`, naming the language the project's meta is
+written in.
+
+Meta is the human-readable prose authored inside the code rather than in a
+message catalogue: `description` on functions and steps, `name`/`title` on
+features and scenarios, step `template` strings, role and persona descriptions.
+It is also the one part of a project the Pikku Console renders back to a human,
+which is the entire reason the field exists — a team whose working language is
+German should be able to read their own Console in German without anything else
+about the project changing.
+
+It defaults to `en`, is validated as a BCP-47 tag through
+`Intl.getCanonicalLocales` (so `de_DE` fails at the line that is wrong rather
+than degrading to "some language" three layers down), and comes back
+canonicalized so `EN-gb` and `en-GB` are one value downstream.
+
+Two things it deliberately does not do, because collapsing them is the bug this
+came from. It never renames anything — identifiers, files, database tables and
+columns stay English whatever it says. And it is not the product's UI language:
+what the app says to its users lives in `messages/<locale>.json`, with
+`active.json`'s `defaultLocale` deciding what a first-time visitor is served,
+while `baseLocale` names the message source and stays `en`.
--- a/.changeset/skills-name-the-three-languages.md
+++ b/.changeset/skills-name-the-three-languages.md
@@ -1,5 +1,5 @@
 ---
-'@pikku/skills': minor
+'@pikku/skills': patch
 ---
 
 Skills now name the three languages a project has, and refuse to let them
--- a/.changeset/skills-name-the-three-languages.md
+++ b/.changeset/skills-name-the-three-languages.md
@@ -1,0 +1,24 @@
+---
+'@pikku/skills': minor
+---
+
+Skills now name the three languages a project has, and refuse to let them
+collapse into one.
+
+An agent building a doctor's portal for a German practice read "the entire UI is
+German" as an instruction about the codebase. It shipped
+`project.inlang/settings.json` with `baseLocale: "de"` and no `en.json` — which
+broke `--add-locale` permanently — alongside RPC functions `getUebersicht` and
+`getPatientendetail`, components `Zeitstrahl` and `AufmerksamkeitStreifen`, and
+database tables `vorgang` and `ereignis`. Nothing in the skills had ever told it
+these were three separate decisions, so it made one.
+
+`pikku-concepts` now carries the canonical statement — identifiers are always
+English and no setting changes that; meta (`description`, `title`, `template`)
+follows `locale` in `pikku.config.json`; the product's UI language lives in the
+message catalogue with `defaultLocale`, never in `baseLocale`. `pikku-build-app`
+§1a asks the question and writes the answer into the config; `pikku-scenario`
+splits step identifiers from step prose and admits where the English-only
+reporter frame still shows through; `pikku-i18n` states why `baseLocale` stays
+`en` and what to set instead. `pikku-build-quick`, `pikku-build-platform`,
+`pikku-feature` and `pikku-fabric` carry the short form.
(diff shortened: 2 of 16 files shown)


yasserf (commented):
Two things before this merges.

**The bumps are `minor`.** `meta-has-a-language-identifiers-do-not.md` sets `'@pikku/cli': minor` and `skills-name-the-three-languages.md` sets `'@pikku/skills': minor`. The packages are on 0.12.120 and 0.12.16 — pre-0.13, everything is `patch`, with minor and major held back until the release that carries the compatibility commitments. The third changeset here already gets this right.

**Nothing reads `locale` yet.** `_getPikkuCLIConfig` does `result.locale = normalizeLocale(result.locale)` and that is where it stops — no codegen path and no Console path in this diff consumes it. Both the changeset and the `PikkuCLIInput` JSDoc describe the capability as present ("a team whose working language is German should be able to read their own Console in German without anything else about the project changing"), when what ships here is a validated, canonicalized config field plus the skills that explain it. Worth either landing the consumer alongside it or wording the release note as the groundwork it is, so nobody sets `locale: "de"` and files a bug when the Console does not change.

The rest reads well. `normalizeLocale` catches the mistake people actually make — `de_DE` — at the line that is wrong, returns the canonical spelling so `EN-gb` and `en-GB` compare equal downstream, and rejects the empty string and the wrong type through the same message. The `app-base-locale-not-english` finding is nicely judged: defaulting `baseLocale` to `en` before comparing, naming both halves of the fix, and adding the re-key note only when the declared locales show the app is already keyed in the other language.

One nit: the two `assert.ok(...)` reformats in `validate.function.test.ts` are prettier churn unrelated to this change.

PR #1718, opened 2026-09-16
knowledge plan set refuses a milestone that already has a plan

yasserf:
`pikku knowledge plan set` validated a plan but did not refuse one when the milestone already had a plan on disk, so the plan could be rewritten to match whatever the build did. The plan is an order: written before the code, fixed afterwards, and `plan defer` is the only way down.

- `runKnowledgePlanSet` returns `ok: false` with a "plan is fixed, use `plan defer`" problem when `readPlan` succeeds for the milestone; an unreadable plan file can still be replaced.
- Two tests: a second `plan set` is refused and leaves the first plan untouched; a corrupt plan file is replaceable.
- `pikku-architect/SKILL.md` says the plan is written once.


yasserf:
Folded into #1717 — one PR for the plan-is-an-order work.

PR #1719, opened 2026-09-16
Scenario suites can reset the local database between scenarios

yasserf:
`scenarios.reset` in `pikku.config.json` makes the runner roll the local database back between scenarios, so a project no longer has to ship a reset of its own.

```json
{
  "scenarios": {
    "reset": { "enabled": true, "keep": ["user", "session", "account"] }
  }
}
```

## What it does

The runner copies every table once, from the migrated and seeded database the suite is pointed at, and replaces the rows from that copy before **every scenario**.

`keep` names the tables it leaves alone. The session tables belong there: an actor signs in once for the whole run, and rolling its session back to a copy taken before that sign-in signs every actor out mid-suite.

### Features get a second layer

A feature's `before` hook is meant to run once and build context its whole group shares, which a per-scenario rollback would delete. So the hook runs against the seed, and what it leaves behind is captured as a second layer. Each of that feature's scenarios then rolls back to the seed *and* the hook's fixtures — isolation stays per scenario, without the hook running again or its work being thrown away. The layer is popped when the feature ends, so the next feature starts from the seed rather than the last one's leftovers.

## Why copy-and-replace, not truncate-and-reseed

A project's dev seed is plain `INSERT`s — that is the point of `pikku db reset` owning it, so a seed never has to defend itself with `ON CONFLICT DO NOTHING`. Replaying it over the tables `keep` held back is a primary key conflict, not a reset.

Foreign keys are suspended for the duration (`PRAGMA foreign_keys` on sqlite, `session_replication_role` on Postgres), because `keep` can hold the parent of a table being replaced and no ordering makes that legal. Postgres sequences are captured and `setval`'d back with the rows: restoring rows without them leaves the next insert claiming an id that already exists.

The copies are dropped at the end of the run — a database left holding them reads as schema drift the next time anything introspects it.

## Why it is in the framework

The alternative it replaces is a project writing its own `testReset` RPC and gating it on an environment variable. That function is in the deploy bundle on every stage, exposed and unauthenticated, and the flag that arms it is one wrong project id away from truncating a live database.

Nothing here reaches a bundle. It is the runner talking to a database file or server on this machine, and it is refused outright when the environment is marked `production`, when `apiUrl` is not a local hostname, and when `NODE_ENV=production`.

## Tests

`packages/cli/src/functions/db/scenario-baseline.test.ts` covers the sqlite path: rows added, removed and changed by a scenario all come back; `keep` tables are untouched; a restore succeeds with both sides of a foreign key emptied; a feature layer survives a scenario rollback and pops back to the seed; `drop()` clears every layer's copies; and production is refused.

## Summary by CodeRabbit

- **New Features**
  - Added optional database reset support for scenario runs.
  - Restores the local database before each scenario while preserving feature setup fixtures.
  - Supports excluding specified tables from resets.
  - Prevents resets against production environments or non-local API URLs.

- **Bug Fixes**
  - Improved scenario failure messages to identify whether setup or database restoration failed.

- **Tests**
  - Added coverage for local URL detection and database baseline restoration behavior.


yasserf:
Addressed the three review findings — all three were real.

**Remote Postgres URL (`scenario.ts:488`).** Correct: `isLocalUrl(env.apiUrl)` said nothing about where the database is, and a local server can be pointed at a remote `postgresUrl`. `captureScenarioBaseline` now refuses a `mode: 'url'` Postgres whose hostname is not loopback, before any client is constructed. Two tests cover it.

**Cleanup outside a `finally` (`scenario.ts:825`).** Correct. The whole post-capture body of `scenarioRun` is now inside `try { … } finally { drop }`, so a throw in `runStore.start`, `browserLifecycle.reset` or the group loop no longer strands the `pikku_scenario_baseline__*` copies. The `drop()` failure still only warns, so it cannot replace the original error.

**Sequences owned by kept tables (`postgres/scenario-baseline.ts:74`).** Correct, and the sharpest of the three. The `pg_sequences` query now joins through `pg_depend` to the owning table and skips any sequence whose owner matches `keep` — winding that counter back while its rows survive the restore is a guaranteed id collision on the next insert.

The `Templates (24, nextjs, yarn)` failure on the previous commit was yarn 4.9.2 dying in its own resolution step (`Error: The \`onCancel\` handler was attached after the promise settled`, then `test-app@workspace:. … doesn't seem to be present in your lockfile`), not this change.

PR #1726, opened 2026-09-17
fix(console): page options outside our chrome, and a scenario run report you can read

yasserf:
Two changes to `@pikku/console`, found together: the second is what you see, the first is what stopped you seeing it.

### `PageOptionsPortal` no longer throws outside `PageOptionsProvider`

`ResizablePanelLayout` moves its left drawer (and side panel) into the phone sheet below 48em, via `PageOptionsPortal`. That portal called `usePageOptions()`, which throws when the provider is absent — and the provider is only mounted by our own `AppLayout`. A host that brings its own chrome and renders our pages directly therefore crashed on every drawer-bearing screen at phone width: scenarios, http, agents, workflows, middleware, the lot.

`useOptionalPageOptions()` returns the context or `null`; the portal skips its registration effect without one and renders its children in place rather than portalling them, so the rail stays where it was declared instead of disappearing. This is the tolerance `usePageOptionsDismiss` already documented for exactly this case — the portal just never got it.

### The scenario run report

Every scenario in a run rendered expanded, and every video artifact rendered its own `<video>`. Because `useScenarioArtifact` fetches the bytes as a blob (an `` cannot carry the console's `Authorization` header), a 37-scenario run downloaded all 34 recordings on open and stacked 34 players down one column, with no filter, grouping or search.

Now:

- **`ScenarioRunTimeline`** — the run as one bar, a segment per scenario sized by duration and coloured by status. Where the time went and where the failures are, before any scrolling; clicking a segment opens that scenario.
- **Scenarios collapse.** Each result is a row (status, name, feature, tags, duration); one is open at a time, defaulting to the first failure. Nothing mounts a recording until you open it.
- **`ScenarioRunPlayer`** — one player for the open scenario, beside its ladder. Steps are buttons; picking one seeks the recording to it. `stepOffsets()` derives each step's offset as a running sum of the durations before it, which is how the recording was laid down.
- `ScenarioArtifactTile` is screenshots only now.

Three new `en` keys. No API change beyond the new exported hook.

## Summary by CodeRabbit

- **New Features**
  - Scenarios and runs now appear in a unified suite view with selectable run lenses.
  - Scenario progress includes live status indicators, timelines, durations, failure details, and step-linked recordings.
  - Feature bars show individual scenario outcomes and counts.
  - Scenario runs record the suite version, repeat attempts, and working-tree state.
  - Embedded hosts can provide page-options state for consistent console rails.
  - Scenario filters support multi-select options.
  - Recordings and artifacts appear in a poster-style rail with step-based seeking and full-size still-image viewing.

- **Bug Fixes**
  - Page options render inline when no provider is available.


yasserf:
Duplicate of #1780: same commits, rebased separately. #1780 carries them plus one docs commit.

PR #1735, opened 2026-09-17
fix(inspector,cli): global middleware survives per-unit builds, and the CLI stops printing stacks

yasserf:
Two independent CLI fixes, raised together because they were found in the same debugging session.

## 1. A global middleware registration can no longer be overwritten

Middleware instance ids were allocated per source file: the per-group index restarted at 0 for every `add*Middleware` call, so the second file to register against a group minted ids the first already owned and replaced its entries in `middleware.instances`. `instanceIds` then listed one key twice while `count` correctly said two.

For `addGlobalMiddleware` that was not cosmetic. Global middleware belongs to no wire group, so the instance map is the only record that its module has to be imported, and per-unit deploy codegen emits its side-effect imports from that map. An app file registering a global middleware was erased by the generated auth scaffold registering its own — the app's module was never imported, its registration never ran, and since auth commonly lives in global middleware **the deployed unit answered every authenticated route with a 401 against a completely clean build log**. Adding any wire registration to the same file hid the bug, because that gave the file a group and a group is imported on its own account.

This is not hypothetical — it took out every authenticated route on a Fabric stage, and only appeared at all because that project had recently moved from per-function to per-service unit grouping.

Ids are now allocated from the ids already present, so N registrations against one group produce N distinct entries whatever order the files are visited in.

The inspector also records each `addGlobalMiddleware` file in `middleware.globalFiles` — a plain set, which no key collision can corrupt. Codegen emits the side-effect imports from it, generates the middleware file when global middleware is a project's only middleware (a session bridge from a package leaves no local definition and used to generate no file at all), and asserts against the text it emitted that every such file is imported, naming the files if not rather than shipping a unit without its auth gate.

## 2. A prompt with nobody to answer it is an error, not a hang

`pikku fabric secrets set NAME` read its value through readline in terminal mode. On a stdin that is not a tty that promise never settles, so the command printed the prompt and then node's "Detected unsettled top-level await" naming a line of pikku's own bin — and under bun it hung outright.

The prompt now reads the first line of stdin when there is no tty, so piping a value works, and refuses with the flag to reach for when stdin is closed or empty. `promptConfirm` gained the same backstop.

Everything else a raw stack could still escape through goes with it: the binary and the generated bootstraps format through `formatCLIError` instead of `console.error(error.message)` and install handlers for what escapes the await, every remaining plain `Error` raised by a fabric command becomes a `PikkuError`, and the watcher and MCP schema loader log their causes at debug level. `--verbose` / `PIKKU_DEBUG` still prints the stack, and an unexpected error still keeps its frames by default.

`packages/cli/src/fabric/lib/errors.test.ts` guards the directory, so a new bare `throw new Error(` in a fabric command fails the suite rather than printing a stack at a user.

## Tests

Added 6 inspector tests and 5 CLI tests covering the id allocation, the
`globalFiles` set, the assertion, and the non-tty prompt paths.

Rebased onto `dfd70193a`. `bun run build:packages` is green, and the suite is
green but for `the shipped surface doc > says what each key it lists is for`,
which fails identically on a clean checkout of main — 807 of 1037 keys, 78%
against a 79% ratchet, the same two numbers with and without this branch. The
undocumented keys it names are core's analytics types; nothing here adds one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Secret prompts now read the first line of piped input and clearly reject empty or closed input instead of hanging.
  - Global middleware is reliably included in generated deployments, including projects with only global middleware.
  - Middleware registrations no longer overwrite one another across files.
  - Missing CLI configuration now provides clearer guidance.

- **CLI Improvements**
  - Errors are presented consistently with appropriate exit codes and optional stack traces.
  - Fabric command failures now provide clearer, structured error handling.
  - Watcher and schema-loading diagnostics are available at debug level.
  - Added shared CLI error-formatting utilities.


yasserf:
Addressed in 3f53998aa — 6 of 7.

**Taken:**

- **`reportFatal` exiting before stderr flushes** (`bin/pikku.ts`, `serialize-channel-cli-client.ts`, `serialize-local-cli-bootstrap.ts`) — correct, and it undercut the point of the commit: stderr is asynchronous when it is a pipe, so the diagnostic was truncated exactly when redirected into a log. The exit now runs from the write callback, with `exitCode` set first so a stream that is already gone still ends the process with the right code.
- **`hasGlobalMiddleware` missing the legacy `global:*` instances** — correct, and an inconsistency inside this PR. `serializeMiddlewareImports` reads `globalFiles ?? []` *and* sweeps `instances` for `global:*`, deliberately, so a state serialized by an older inspector keeps working. The gate that decides whether that serializer runs read `globalFiles` alone, dropping those older states before it ever saw them, and would have thrown on `globalFiles` being undefined.
- **No command-level coverage for the global-only path** — added `pikku-command-middleware.test.ts`. Its legacy-instances case fails against the old gate, so it is testing the gate rather than the serializer under it.
- **`pikku.config.json` in the not-found message** — correct; the search accepts `.ts`, `.js` and `.json`, so anyone with a `.ts` config was sent looking for a file they were right not to have.

**Not taken:**

- **Treating `process.env.CI` as non-interactive in `promptSecret` / `promptConfirm`.** `stdin.isTTY` is the accurate signal for "nobody can answer this", and it already covers every CI runner that does not allocate a pseudo-tty, which is the normal case. The proposal only changes behaviour where stdin *is* a tty, and there it makes a real terminal refuse to prompt because an unrelated environment variable happens to be set — `CI` is set by a lot of things, including some interactive shells and devcontainers. That is a regression for a human at a keyboard in exchange for a narrower hang than the one this PR fixes. Happy to revisit if there is a concrete runner where stdin is a tty and the prompt hangs.

Full suite green but for `the shipped surface doc > says what each key it lists is for`, which fails identically on a clean checkout of main (807 of 1037 keys, 78% against a 79% ratchet) and is unrelated to this branch.

PR #1758, opened 2026-09-20
feat(fabric): --organization on link and init

yasserf:
## What

`pikku fabric link` and `pikku fabric init` take `--organization`, naming the organization to import into by **slug**, **display name** or **id**.

```bash
pikku fabric link --organization vlandor
pikku fabric init https://github.com/acme/app --organization 2222…
```

## Why

Both commands imported into whichever organization the CLI session happened to be scoped to, with no way to say otherwise short of logging in again. If you belong to more than one org, a repo owned by the company lands in your personal org — or, for a `github.com` repo, refuses with

> Connect the GitHub account "vlandor" before importing this repository

naming an owner the session's org has never been linked to, with no flag to point at the org that *is* linked.

## How

`importProject` already read `input.organizationId ?? session?.orgId`, so this is CLI-side only — no API change. A new `lib/organization.ts` resolves the flag: a UUID passes straight through, anything else is matched case-insensitively against `slug` and `name` from `listMyOrganizations`. An unknown name refuses and lists the orgs you belong to; an ambiguous one refuses with the ids to choose from. Omitting the flag sends no `organizationId` at all, so the existing session behaviour is untouched and no lookup is made.

Membership is still enforced server-side — `importProject` carries `permissions: { member: [canAccessOrgFromInputOrSession] }` — so the flag lets you choose among your own orgs, not reach into someone else's.

## PR checklist

- [x] unit tests — `packages/cli/src/fabric/lib/organization.test.ts`, 6 cases covering the no-flag, id, slug, name, unknown and ambiguous paths
- [x] docs — `packages/skills/skills/pikku-fabric/SKILL.md`, in the Deploy section beside `init`
- [x] changeset — `.changeset/fabric-link-organization-flag.md` (patch)
- [ ] **verifier** — skipped. The resolver is a pure function over one RPC response and is covered directly by unit tests; there is no wiring surface for a verifier to assert against.
- [ ] **e2e** — skipped. Exercising it end to end needs a session that belongs to two organizations, which the e2e fixture does not set up.
- [ ] **knowledge** — nothing to record; the behaviour is documented in the skill above.

Flagging the three skips explicitly rather than omitting them — say the word if you want the e2e fixture extended to a second org and I will do that instead.

## Summary by CodeRabbit

* **New Features**
  * `pikku fabric init` and `pikku fabric link` now support selecting a target organization with `--organization`.
  * Organizations can be specified by slug, display name, or ID.
  * Without this option, imports continue to use the organization associated with the current session.

* **Documentation**
  * Added guidance on organization access requirements and GitHub repository imports.


yasserf:
Addressed in 188a55dc7.

**1. `link.function.ts:126` — resolve before provisioning. Fixed.** Good catch, and a real one. With `--gitea` and no `origin`, `createRemote` creates a repository on Fabric's git server and pushes to it, so resolving the organization afterwards meant a typo cost the user a repo they then had to go and delete. It now resolves immediately after the `hasCommits` / `isWorkingTreeClean` checks, which sit there for exactly that reason.

**2. `organization.test.ts:57` — case-insensitive slug. Fixed.** Input is now `VlAnDoR`, and the test name says so. The name case was already covered; the slug case was passing for the wrong reason.

**3. Command-level tests asserting the `importProject` payload — skipping, with a reason.**

`functions/` has no test harness today — every test in this package lives in `lib/`, because the command functions reach straight for `resolveApiContext`, `getFabricRPC` and the git helpers with no seam to stub. Covering `FabricLink` would mean building that harness (module mocks for HTTP, config and git) inside a PR that adds one flag, and the harness would then be the bulk of the diff.

What is left uncovered is the forwarding itself — one spread, `...(organizationId ? { organizationId } : {})`, on each command — while the resolver behind it has 6 cases. That is a thin enough seam that I would rather not pay for it here.

Worth doing as its own change, since the ordering bug you found above is precisely the kind of thing a command-level test would have caught. Happy to open that separately if a maintainer agrees it is wanted.

PR #1770, opened 2026-09-20
feat(addon-typesafe): TypeSafe System One as an addon

yasserf:
Adds `@pikku/addon-typesafe` — [TypeSafe System One](https://console.typesafe.ai) as a Pikku addon. It returns typed judgments and calibrated probabilities rather than generated text.

Fabric had this as a service on an unmerged branch, wired into one enricher. The judgments are not fabric-specific, so the addon is the right home: any project reaches them with a `ref()`.

## Functions

```
typesafe:classifyTask   { task, context? }
  -> { plan:  { needed, noul },
       shape: { choice: 'function' | 'workflow' | 'agent', confidence, probabilities } }

typesafe:ask            { state, questions } -> { answers }
```

Both `classifyTask` judgments are asked in one request against the same state, so the shape is judged by an evaluator that has already weighed how well specified the task is. There is no `unknown` option in `shape` — confidence is the don't-know channel, and a flat distribution says "underspecified", which is itself a reason to plan.

`ask` is the unopinionated door: `noul` for one 0–1 judgment, `choice` for named options, batched so related questions are one call and mutually consistent.

## Notes

- **Neither function carries `expose: true`.** They are reachable through `ref()` as an agent tool, a workflow step or a wired route, where the caller's own gate lives. An anonymous `POST /rpc` onto a metered third-party API is not a default worth shipping, and PKU574 says so.
- **`TYPESAFE_API_KEY` is declared with `allowedHosts: ['api.typesafe.ai']`** and read as a literal, so the addon is scoped to exactly that secret and needs no `secretGrants`. Consumers must not hand it `globalSecrets` — it makes outbound requests.
- TypeSafe bills directly, so its spend does not appear in a consuming app's LLM-proxy accounting. `SystemOneService.askWithUsage` (exported from `@pikku/addon-typesafe/service`) returns token counts for apps that meter their own usage.

`pikku all` and `tsc --noEmit` are clean.


yasserf:
Wrong repo — addons live in pikkujs/addons. Rebuilding it there.

PR #1779, opened 2026-09-21
fix(inspector): make generated type aliases deterministic

yasserf:
## The bug

A type name that collides across files is disambiguated with a random suffix (`types-map.ts:16`):

```ts
const uniqueName = `${originalName}_${Math.random().toString(36).substring(7)}`
```

So codegen output is different on every run. Three consecutive `pikku all` runs on `e2e`:

```
run1 agent=8ca562ac… 61389 bytes
run2 agent=d6f1c6c8… 61390 bytes
run3 agent=61bad7f6… 61390 bytes
```

```
RecordedEvent as RecordedEvent_acvwoj   ← run 1
RecordedEvent as RecordedEvent_febca8   ← run 2
RecordedEvent as RecordedEvent_ih2h3d   ← run 3
```

(The ±1 byte is 5- vs 6-character random suffixes.)

## Why it costs 20 seconds

Those aliases land in `.pikku/agent/pikku-agent-map.gen.d.ts` and `.pikku/workflow/pikku-workflow-map.gen.d.ts`, which are both inside the schema generator's dependency closure. `fingerprintSchemaDeps` records their mtime and size; `schemaDepsUnchanged` then correctly reports a change, and the whole TS schema cache is discarded.

Instrumenting a single `pikku all` shows it happening twice, mid-run:

```
SCHEMA-DISKHIT
WRITE .pikku/agent/pikku-agent-map.gen.d.ts
SCHEMA-GENERATE                                 ← 674 types
WRITE .pikku/rpc/pikku-rpc-wirings-map.internal.gen.d.ts
WRITE .pikku/rpc-map.gen.d.ts
WRITE .pikku/workflow/pikku-workflow-map.gen.d.ts
SCHEMA-GENERATE                                 ← 674 types again
WRITE .pikku/http-map.gen.d.ts
SCHEMA-MEMHIT
```

Each regeneration produces byte-identical output — `md5=75058a4d3ac05c412d9d2367475aac9b`, 674 keys, every time. The cache was working exactly as designed; it was being fed a moving target.

CPU profiling attributes 67% of a full `pikku all` to `ts-json-schema-generator` (its `inspectNode`/`getFullName` dragging TypeScript's symbol-name printing behind them), which is what this work is.

## The fix

Derive the suffix from the declaring file's path. A counter handles the case where one path legitimately needs two aliases; `exists(originalName, path)` already dedupes that pair before `addUniqueType` is reached, so the only caller can't hit it.

## Measured

Same machine, same state, on `e2e`:

| | before | after |
|---|---|---|
| warm `pikku all` | 23s | **3s** |
| schema generations (warm) | 2 | **0** |
| cold `pikku all` | 23s | 23s |
| output stable across runs | no | **yes** |

CI builds once cold and is warm for everything after, so this applies to every `pikku all` after the first in a job.

Cold is unchanged and needs a separate change: schema generation happens inside `inspect()`, so each re-inspection regenerates. Making `state.schemas` lazy would collapse a first build's three passes to one.

## Validation

- 869 inspector tests pass, 0 fail
- `benchmarks/bench-codegen-perf.ts` passes every gate (3 passes, 399MB peak, 2039ms)
- `tsc --noEmit` clean on `@pikku/inspector`
- No committed `.gen.` file in the repo carries these suffixes, so there is no one-time churn to land

## Summary by CodeRabbit

- **Bug Fixes**
  - Generated type names are now consistent across repeated runs, preventing unnecessary changes to generated files.
  - Schema generation now reuses cached results reliably instead of triggering a full regeneration when names previously changed unexpectedly.
  - Improved handling of type-name collisions ensures distinct types remain uniquely identifiable across source files.


yasserf:
Addressed in df636b8 — but by removing the order-dependent branch rather than testing around it.

The finding is right: if two paths sharing an `originalName` collided on the six-character digest, the base alias went to whichever was inserted first and `_1` to the second, so a traversal-order change would swap them. That is the same nondeterminism this PR exists to remove, just rarer.

Rather than assert the counter behaves, the counter is gone. The digest widens to 12 hex characters and the alias is now a pure function of the path, with no branch on the map's current contents:

```ts
const digest = createHash('sha256').update(path).digest('hex').slice(0, 12)
const uniqueName = `${originalName}_${digest}`
```

Order-independence is now structural, so there is nothing left for insertion order to influence.

Added `packages/inspector/src/types-map.test.ts` with the requested verifier plus three siblings:

- the same path always yields the same alias
- **insertion order does not change which alias a path gets** — inserts two paths forward and reversed, asserts each keeps its alias
- distinct paths sharing a name get distinct aliases
- the alias is what `exists()` hands back for that name and path

873 inspector tests pass, 0 fail. On `e2e`, warm `pikku all` stays at 3–4s with a stable output hash across runs.

